PR #2, opened 2024-03-27
Added necessary migrate command for freshly cloned repos.

mattash:
Yay, my first PR!

I had to run migrate before I could create super use, so I added it to the readme file.


andylitalo (approved):
Thanks for the addition--good catch!

PR #4, opened 2024-04-06
[feature] add participant count to fast endpoint

andylitalo:
**Feature**

Modified existing fast endpoint to also return the "participant count," i.e., the number of people participating in the fast (including the user). Only required adding participant count to the FastSerializer and defining a method to count the number of participants in a fast.

**Additional Changes**

* Refactored retrieval of fast from request
* Updated README

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,10 +78,13 @@ user to participate in
 
 We are currently testing one endpoint:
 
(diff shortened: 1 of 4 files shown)


andylitalo:
Looks good--manually tested

PR #5, opened 2024-04-18
Maintenance/prep for deploy

mattash:
These changes make the bahk app compatible with Heroku. Eventually will need to incorporate CI review compatibility.


bahk/settings.py
```diff
@@ -119,6 +128,10 @@
 
 STATIC_URL = 'static/'
 
+STATIC_ROOT = os.path.join(BASE_DIR, 'staticfiles')
+
+STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
```
andylitalo, line 133: Is there a reason for using this format to specify the static files storage method? According to [this](https://whitenoise.readthedocs.io/en/stable/django.html#add-compression-and-caching-support), for Django >= 4.2 (`requirements.txt` specifies `django==4.2.11`), the following format is preferred:
```
STORAGES = {
    # ...
    "staticfiles": {
        "BACKEND": "whitenoise.storage.CompressedManifestStaticFilesStorage",
    },
}
```

andylitalo (commented):
One nit, but otherwise LGTM!

bahk/settings.py
mattash: You are right about this. When you run a test it shoots a deprecation notice. Can't remember if I tried it and got stuck or was lazy. Let's resolve this before merging.

mattash: So I looked into it and to use the white noise module which is required for virtual hosts, the old syntax is required. We could switch to using Amazon S3 for static file service, but I dont think its necessary because its an api app. Anyway, I think we should suffer the warning for now.

andylitalo: Sounds good to me--compatibility much preferred over reducing warning messages.

PR #7, opened 2024-05-07
Feature/fast description

mattash:
Adds a description field to Fast model


andylitalo:
Looks great! I thought it would be nice and within scope to display the description in the fast serializer and fast admin page (see latest commit)

PR #9, opened 2024-05-09
Adds JWT

mattash:
Anticipating that we will be building a client app, using something like react or svelte or mobile native, we will need json based token authentication.


andylitalo:
LGTM! Thanks for setting up groundwork for using web tokens

PR #11, opened 2024-05-11
Feature/culmination fast field

andylitalo:
**Feature**

* Adds culmination feast and feast date
* Adds countdown to culmination feast in fast serializer (or countdown until first day after fast)

Also updates pytest `test_endpoints.py` to pass with new serializer.


andylitalo:
I've tested and it passes automated tests, so I'll merge

PR #17, opened 2024-06-06
Bugfix/s3 media

mattash:
s3 hosted files were not dynamically resizable in views. This resolves that issue.


hub/models.py
```diff
@@ -49,25 +47,9 @@ class Profile(models.Model):
...
-        img = Image.open(image)
-        img = ImageOps.exif_transpose(img)
-        img.thumbnail(size, Image.LANCZOS)
-        buffer = BytesIO()
-        img_format = img.format if img.format else 'JPEG'
-        img.save(buffer, format=img_format)
-        content_type, _ = mimetypes.guess_type(image.name)
-        return InMemoryUploadedFile(buffer, 'ImageField', image.name, content_type, buffer.tell(), None)
+    location = models.CharField(max_length=100, blank=True, null=True) 
+    profile_image = models.ImageField(upload_to='profile_images/originals/')
```
andylitalo, line 51: It looks like there are two profile_image fields. If this is intended, could you change the name of one of them?

andylitalo:
Could use more testing (for PNG uploads, which are not currently supported--just jpg). Leaving for separate effort.

Note: merged migrations for consolidation--please migrate back to migration 0007 (`python manage.py migrate hub 0007`) before migrating to the new migration (0008)

PR #29, opened 2024-06-25
Should resolve rotation of certain uploaded images after being processed

### LINKED ISSUE #24: Profile picture appears rotated 90 degree (horizontal) when displayed as avatar
Live and Pray

First avatar thumbnail is horizontal, but appears vertical in "Profile" section.

andylitalo:
Fixes #24

Uses imagekit library to transpose images based on metadata instead of aspect ratio so profile images that are taller than they are wide are not rotated upon display (mobile issue only).

Tested on iPhone on local server--bug fixed!

### DIFF
--- a/hub/models.py
+++ b/hub/models.py
@@ -4,6 +4,8 @@ from django.db import models
 from django.db.models import constraints
 from imagekit.models import ImageSpecField
 from imagekit.processors import ResizeToFill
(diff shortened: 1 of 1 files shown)


mattash (commented):
Looks great!

mattash (approved):
still looks good

PR #30, opened 2024-06-26
Feature/email notifications

mattash:
In order to test locally, you have to

1. Update modules
2. Run `python manage.py migrate`
3. Install Redis on your local machine
4. Run `redis-server`
5. Update your .env to include
    1. REDIS_URL=redis://localhost:6379/0
    2. REDIS_TLS_URL=redis://localhost:6379/0
    3. MAILGUN_API_KEY=[can send privately]
    4. MAILGUN_DOMAIN='mg.armenian.faith'
7. Run `celery -A bahk worker --loglevel=info`
8. From `manage.py shell` you can run the following: 

```
from hub.tasks import send_fast_reminder_task 
send_fast_reminder_task()
```
This will trigger the task that sends email reminders to all users

_or_

```
from hub.utils import send_fast_reminders
send_fast_reminders()
```
This will bypass the task system, and just send the email reminders to all users

To just test if emails are working:
```
from hub.utils import test_email
test_email()
```
You should adjust the could in hub/utils.py to send test emails to the desired email address


bahk/settings.py
```diff
@@ -187,10 +189,33 @@
...
+EMAIL_USE_TLS = True
+EMAIL_HOST_USER = 'postmaster@' + config('MAILGUN_DOMAIN')
+EMAIL_HOST_PASSWORD = config('MAILGUN_API_KEY')
+ANYMAIL = {
+    "MAILGUN_API_KEY": config('MAILGUN_API_KEY'),
+    "MAILGUN_SENDER_DOMAIN": config('MAILGUN_DOMAIN')  
+}
+EMAIL_BACKEND = "anymail.backends.mailgun.EmailBackend" 
+DEFAULT_FROM_EMAIL = "bahk@gmail.com"
+# SERVER_EMAIL = "your-sender-email@example.com"
```
andylitalo, line 212: Should this be an env var defined by the host server?

hub/urls.py
```diff
@@ -26,6 +26,7 @@
     path("web/", views.home, name="web_home"),
     path('add_fast_to_profile/<int:fast_id>/', views.add_fast_to_profile, name='add_fast_to_profile'),
     path('remove_fast_from_profile/<int:fast_id>/', views.remove_fast_from_profile, name='remove_fast_from_profile'),
+    path('test-email/', views.test_email_view, name='test_email')
```
andylitalo, line 29: could this url segment be `'test_email'` (with underscore) for consistency?

hub/utils.py
```diff
@@ -1 +1,63 @@
...
+            email.send()
+            logger.info(f'Fast reminder sent to {profile.user.email} for {next_fast.name}')
+
+def test_email():
+    try:
+        send_mail(
+            'Test Email',
+            'This is a test email sent from Celery.',
+            settings.EMAIL_HOST_USER,  # Replace with your sender email
+            ['matt.ash@gmail.com'],  # Replace with the recipient email
```
andylitalo, line 57: Maybe consider making this an env var since this is a public repo? :)

andylitalo (changes requested):
A few nits, but passed the smoke test (`test_email()`)--would be good to test the fast reminders in a staging environment before deploying if you haven't already.

hub/utils.py
mattash: Yes. We can call it TEST_EMAIL or something.

hub/urls.py
mattash: That's fine. If we change here, does it need to be changed anywhere else?

andylitalo (approved):
Remaining testing will be performed on prod (like a staging environment since not many users)

PR #31, opened 2024-06-28
Feature/accent font

mattash:
Adds Stoke as font for Fast titles on main page. Other very minor UI tweaks.


andylitalo (approved):
LGTM! I like the new font and the feature to see user names on hover over the avatars.

PR #32, opened 2024-06-29
Feature/test user tools

mattash:
This adds a new module for the app called "app_management" that can be used to post changelog updates for users to learn more about what changes are being pushed to the app. 

It also adds a button for users to provide feedback and submit bug reports.


hub/templates/changelog.html
```diff
@@ -0,0 +1,19 @@
+{% extends "base_generic.html" %}
+
+{% block content %}
+<div class="row w-100 p-3 mx-3 mb-3 rounded-bottom shadow bg-light text-dark">
+
+  <h2>Changelog Updates</h2>
+  {% for changelog in changelogs %}
```
andylitalo, line 7: In the edge case where there are no changelogs, this loop throws an error. Should we just skip the loop with an `if` block if there are no changelogs, or is the error good as a reminder to populate the changelog field?

andylitalo (changes requested):
One small question that might need a code change for displaying changelogs when there is none in the database

hub/templates/changelog.html
mattash: Good point. I can handle that.

mattash: fixed here: https://github.com/surp-hovhannes/bahk/pull/32/commits/dfe3ad7e055437f2027c97288fc0e74828a095da

andylitalo (approved):
LGTM!

Issue #26, opened 2024-06-19
Allow user to click on fast card image to open "Read More" modal with fast description


andylitalo:
Vague attempt at this [here](https://github.com/surp-hovhannes/bahk/tree/feature/click-fast-card-opens-modal)

mattash:
@andylitalo I worked on this a little and its in a working state. I'm pushing the branch with a PR.

andylitalo:
I'm going to close this since it's working well enough on staging.

PR #33, opened 2024-06-30
Feature/click fast card opens modal

mattash:
Checkout #26


andylitalo:
Right now, clicking anywhere inside the image on the fast card opens the modal, including the "Join Fast" and "Leave Fast" buttons, so when clicking either button, the modal starts to pop up before the user is redirected. I think we'll need javascript to exclude them from the clickable region (see [here](https://stackoverflow.com/questions/70861330/exclude-child-components-from-div-onclick)).

Holding of from merging for now.

mattash:
Good catch. I have a good idea how to fix it.

mattash:
I resolved previous comment in https://github.com/surp-hovhannes/bahk/pull/33/commits/9bb659ce86bdbaf8358c729bb70629e7dff40faf

hub/templates/components/fast_card.html
```diff
@@ -1,5 +1,5 @@
 <div class="col-lg-4 col-md-6 col-sm-12 text-center">
-    <div class="card mb-3" style="background-image: url('{{ fast.image }}'); background-size: cover;">
+    <div class="card mb-3" data-bs-target="#{{ fast.modal_id }}" data-bs-toggle="modal" onclick="if(event.target.tagName !== 'BUTTON' && !event.target.closest('form')) { location.href='#'; }" style="background-image: url('{{ fast.image }}'); background-size: cover; cursor: pointer;">
```
andylitalo, line 2: for some reason the modal still starts to open when I click the Join Fast or Leave Fast buttons, even though you're excluding buttons from the event target tag names...

andylitalo (changes requested):
Looks like it should work but modal still opens upon button clicks in local testing :(

andylitalo:
I'm going to approve since I think giving people the ability to learn about the fast by clicking on the card is more valuable than flawless UX :)

andylitalo (approved):

PR #34, opened 2024-07-01
Fixes existing tests

mattash:
Test suite hasn't worked for awhile. fixed some of the existing tests, but would be worthwhile to build more tests in the future


andylitalo:
I was thinking the same but was going to try restructuring the orm first.

PR #36, opened 2024-08-21
adds app.json to streamline heroku instances

mattash:
Heroku allows for schema to streamline the process of instantiating a new instance. This will allow it to auto generate new "review apps" to see how they behave in the production environment.


mattash:
this is incomplete because it doesn't define testing environment. Will close, and reopen.

PR #35, opened 2024-08-21
Maintenance/simplify db schema

andylitalo:
Replaces m2m relationship with a foreign key between `Fast` and `Day` to keep things simple in the future.

Nothing breaks on local deployment--Der Hayr, could you make a second heroku instance to try this out with the live database? GCP was indeed tougher than what I had thought...

I'll redo the tests after putting this at rest.

PR #37, opened 2024-08-21
Feature/heroku app json

mattash:
Adds app.json to streamline the creation of instances on heroku for the purpose of reviewing PRs in a production environment.

Note: the parameter for test script in app.json is set to "python manage.py test" when the actual testing suite is at "pytest". However, since our testing suite is broken this would cause a nuisance with CI testing, and prevent deployment. When the testing suite is resolved this setting should be fixed.


mattash:
@andylitalo this is in response to PR #35. Roll this update into that PR, and I can more easily create a review version of the app on Heroku that will pull in the existing production DB for testing. PS: i could have manually configured that review app, but this way it'll be easier in the future to test in production environment.

andylitalo (approved):
LGTM! Cool config-as-code feature

PR #40, opened 2024-08-23
Fixes #38 by adding missing JS source

### LINKED ISSUE #38: Feedback form does not work
Clicking on the feedback link in the footer does nothing.

mattash:


andylitalo (approved):
Don't know enough js or html to know what's going on, but it works, so LGTM!

PR #43, opened 2024-08-23
Bugfix/upcoming fast emails query broken

### LINKED ISSUE #42: Recent fast/day model changes break upcoming fast email notifications

mattash:
Fixes #42

### DIFF
--- a/hub/utils.py
+++ b/hub/utils.py
@@ -6,7 +6,7 @@ from django.utils.html import strip_tags
 from hub.models import Profile, Fast, Day
 from datetime import datetime, timedelta
 from .serializers import FastSerializer
-from django.db.models import OuterRef, Subquery
+from django.db.models import OuterRef, Subquery, Q
 
 import logging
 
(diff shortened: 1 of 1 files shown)


hub/utils.py
```diff
@@ -16,21 +16,30 @@
 def send_fast_reminders():
     today = datetime.today().date()
     three_days_from_now = today + timedelta(days=3)
-    
+
+    # Subquery to find the next fast for each profile, excluding those with "Friday Fasts" or "Wednesday Fasts" in the name
+    # TODO: Find a better way to handle weekly fasts
```
andylitalo, line 21: would it help to add a field to a fast marking it as a weekly vs consecutive fast?

andylitalo (approved):
LGTM! Looks like it fixes the issues from the migration and catches exceptions better

hub/utils.py
mattash: Yes, I think so. We could also have two options for notifications in profile. Opting in to annual fasts, and opting in to weekly fasts. Maybe some people want the weekly reminder.

PR #44, opened 2024-08-26
[bugfix] add app and permissions to allow cors headers

andylitalo:
Necessary so axios can send REST API requests.

Still needs to be tested.


bahk/settings.py
```diff
@@ -221,4 +223,9 @@
 if IS_PRODUCTION:
     print("Running tests in production environment")
 else:
-    print("Running tests in development environment")
\ No newline at end of file
+    print("Running tests in development environment")
+
+# handling CORS headers
+CORS_ORIGIN_ALLOW_ALL = config('CORS_ORIGIN_ALLOW_ALL', default=False, cast=bool)
```
mattash, line 229: allow all is set to false by default because this is a potential security hazard. Can be set to All via env vars for testing purposes, but shouldn't be default.

bahk/settings.py
```diff
@@ -221,4 +223,9 @@
...
+CORS_ALLOWED_ORIGINS = config('CORS_ALLOWED_ORIGINS', default='', cast=Csv())
```
mattash, line 230: specific origins can be defined in env vars. For dev purposes the setting CORS_ALLOWED_ORIGINS='http://localhost,http://127.0.01' will do the trick.

mattash (approved):
The functionality was tested using the deployed review app, and basic local react api client.

PR #45, opened 2024-08-26
Feature/enhanced api endpoints

mattash:
- Refactored views by splitting views file into more focused modules (profile.py, fast.py, day.py, etc.), improving readability and maintainability.
- Centralized all API endpoints under the /api/ root path for a more consistent and organized API structure.
- Added multiple endpoints that should expose necessary data and action to clients.
- Updated existing serializers and added new serializers to support expanded API.

API endpoints including authorization have been tested locally using an API client.


hub/serializers.py
```diff
@@ -47,15 +52,16 @@ def get_church(self, obj):
...
-        # test if request is present otherwise return False
-        if 'request' not in self.context:
-            return False
-        else:
-            request = self.context.get('request')
-            if request and hasattr(request, 'user'):
-                return request.user.profile.fasts.filter(id=obj.id).exists()
+    def get_joined(self, obj):
+        # Check if 'request' is present in the context
+        request = self.context.get('request', None)
```
andylitalo, line 57: nit: `None` is the default return value for `.get()`--could condense to
```
request = self.context.get('request')
```

hub/serializers.py
```diff
@@ -47,15 +52,16 @@ def get_church(self, obj):
...
+        
+        # If there's no request or the user is not authenticated, return False
+        if not request or not request.user.is_authenticated:
```
andylitalo, line 60: nit: prefer `request is None` over `not request`

hub/serializers.py
```diff
@@ -99,3 +117,22 @@ def get_next_fast_date(self, obj):
...
+    class Meta:
+        model = models.Day
+        fields = ['id', 'date', 'fast', 'church']
+
+class ParticipantSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = models.Profile
+        fields = ['id', 'user', 'profile_image', 'location'] 
+
+    user = serializers.CharField(source='user.username')  # If you want to include the username instead of the user object
```
andylitalo, line 138: What if we collect first names instead of user names to display for participants? We could use emails for login.

hub/views/fast.py
```diff
@@ -0,0 +1,286 @@
...
+
+    def get(self, request, fast_id):
+        # Retrieve the fast based on the fast_id provided in the URL
+        current_fast = Fast.objects.filter(id=fast_id).first()
+
+        if not current_fast:
+            return response.Response({"detail": "Fast not found."}, status=404)
+
+        # Retrieve up to 6 participants for the specified fast
+        other_participants = current_fast.profiles.all()[:6]
```
andylitalo, line 193: Could we break the number of participants to show out into a constants file? This would allow for easier access in case we want to change/specify different numbers for mobile vs. web.

hub/serializers.py
andylitalo: We can leave this out of scope for this PR--I'll note it as an issue

andylitalo (approved):
Made some small cosmetic changes
Left idea for using first names instead of user names out of scope;
Otherwise LGTM! Really clean and thorough now! Learned a lot from reading through this.

andylitalo:
I approved, but it would be good to test once more since I made some (hopefully just) cosmetic changes. Feel free to merge afterwards.

hub/serializers.py
mattash: Noted. I also was wondering if making usernames accessible via API is problematic. Social networks do that often, but it doesn't pass a smell test for me here. It would make sense to switch to emails as logins, and collect names. I'm not sure we can specify first or just say name, and not that its public... Anyway a matter for another day.

mattash:
Tested live review app using api client. everything seems to check out.

PR #59, opened 2024-09-25
Church Serializer Needs to return ID

### LINKED ISSUE #54: Church Serializer Needs to return ID
We want this to return church id too, so that native client can use it in dropdowns so that users can select their church.

https://github.com/surp-hovhannes/bahk/blob/28247bf3dbbf88977cadcf79dfc97cfc1ef84a90/hub/serializers.py#L39

mattash:
Fixes #54


andylitalo (approved):
LGTM!

PR #62, opened 2024-09-26
Need API Endpoint for registering new users

### LINKED ISSUE #57: Need API Endpoint for registering new users
There is no way to add users via the client now. Need a Serializers, View, and Url route for a new user registration.

mattash:
Fixes #57

client form must submit username, email, password and confirmation password, as well as select their church. If form is successfully submitted it returns user_id, username, email, and church, as well as auth tokens so that the user can be automatically logged in on the client.


andylitalo (approved):
LGTM! Thanks for whipping this up so fast

PR #65, opened 2024-10-02
fixes profile image thumbnail needs to be accessible via profile api …

### LINKED ISSUE #64: profile image thumbnail needs to be accessible via profile api reqest

mattash:
fixes #64


andylitalo (approved):
Nice, helpful having the thumbnail  readily accessible

PR #66, opened 2024-10-08
Adds fast stats endpoint

mattash:
This returns some interesting data about a users participation in fasts that can be displayed on their profile page.


andylitalo (approved):
Cool idea! Looks great

PR #67, opened 2024-10-09
thumbnails added to participant list serializer

mattash:
when generating a list of participants for a given fast, this now also returns the thumbnail. this makes participants lists much faster loading in the client.


andylitalo (approved):
Looks good

PR #68, opened 2024-10-19
Feature/admin show year with fast

andylitalo:
**Features**

- Add `year` field to `Fast` model, which is automatically updated based on the year of the first day of the fast upon saving
- Include the year in the uniqueness constraint to allow for duplicate fasts on different years
- Append year to string representation (e.g, str(lent) -> `Great Lent (2025)`)
- Also some hyperlinking on the admin page to ease editing

**Testing** (I performed this locally, but if you have time to test again I'd feel more confident)

- Click the change form for a fast and save. *Expected outcome:* the year should be automatically updated to match the year of the first day of the fast
- On the admin page, check that the fast with a year shows the year next to the fast name in parentheses
- Check that the new hyperlinks on the change list page of each model work properly
- Check that an identical fast can be created on different years, but not on the same year


mattash (approved):
Looks great! Do you think that the serializer should include the year field? I'm not sure. I'm guessing the user will only be able to see a years worth of fasts so they won't have to differentiate between fasts across years.

andylitalo:
Yeah, I only wanted to display the year to admins and use it to differentiate fasts by year. The frontend shouldn't use the year.

PR #70, opened 2024-10-20
Feature/admin duplicate fast for next year

andylitalo:
**Feature**

Button added to change fast detail page that auto-populates a create fast with dates form using the current fasts properties (except for the year and dates).

**Testing**

Try duplicating a fast! Check that all fields are transferred over except for dates. The year should be automatically calculated based on the days.


mattash (approved):
Tested and works as expected. Awesome!

PR #73, opened 2024-10-31
Adds settings required for celery worker to work on heroku

mattash:
These new settings resolve the issue causing fast notifications to not be sent out, or at least resolves one of the issues. Testing is in the form of checking Heroku logs to make sure that the worker server comes online, where as it was crashing before.

### DIFF
--- a/bahk/celery.py
+++ b/bahk/celery.py
@@ -13,19 +13,10 @@ app.config_from_object('django.conf:settings', namespace='CELERY')
 
 # Configure SSL settings for Redis if using rediss://
 broker_url = os.environ.get('REDIS_URL', 'redis://localhost:6379/0')
-if broker_url.startswith('rediss://'):
-    app.conf.broker_transport_options = {
-        'ssl_cert_reqs': CERT_NONE,
-        'ssl_ca_certs': None,
-        'ssl_certfile': None,
-        'ssl_keyfile': None,
-    }
-    app.conf.redis_backend_use_ssl = {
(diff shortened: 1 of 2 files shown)


bahk/celery.py
```diff
@@ -2,14 +2,35 @@
...
+# Configure SSL settings for Redis if using rediss://
+broker_url = os.environ.get('REDIS_URL', 'redis://localhost:6379/0')
+if broker_url.startswith('rediss://'):
+    app.conf.broker_transport_options = {
+        'ssl_cert_reqs': CERT_NONE,
+        'ssl_ca_certs': None,
+        'ssl_certfile': None,
+        'ssl_keyfile': None,
+    }
+    app.conf.redis_backend_use_ssl = {
```
andylitalo, line 23: It looks like we set the REDIS_BACKEND_USE_SSL in the settings that are already loaded into the config for this app. Do we need to load them again through this syntax here, too?

andylitalo (commented):
Does this fix issues with REDIS trying to find a certificate? Looks good! Just one comment about possible redundancy.

bahk/celery.py
mattash: I removed the duplicated code from the celery.py file. An initial test on the deployment shows that the worker server is stable. Lets merge it and see how it goes.

andylitalo (approved):
Thanks for the changes--glad to hear the code removal didn't affect functionality (yet...)

PR #74, opened 2024-11-03
Bugfix/count results fasts endpoint

mattash:
Adding more fasts for 2025, I realized that the current fast list endpoint only returns 10 results and paginates the rest. This returns results that +/- 6 months from the current date.

To test locally, use an api client, or visit http://localhost:8000/api/fasts/. Append the URL with ?start_date=YYYY-MM-DD and ?end_date=YYYY-MM-DD to test optional params.


andylitalo (approved):
Added a small parenthetical in the docstring to clarify that a fast is included if any day is in the date range. LGTM!

PR #75, opened 2024-11-03
draft of simple endpoint to scrape sacredtradition.am for today's readings

andylitalo:
Assuming sacredtradition.am doesn't change their webpage, we can scrape the daily readings. Here, we add an endpoint to get today's daily readings in a JSON.

Still a draft designed to get us thinking about how we might display or use the readings in the frontend.

**Testing**
Run locally and curl endpoint `/hub/readings/`, e.g.,
```
curl http://localhost:8000/hub/readings/
```


mattash:
This is great, @andylitalo! It's been on my mind to do something like this. I changed the method to a generic API class, and made it possible to query a specific date if desired. Also, I included caching so that the app doesn't have to re-scrape the website when re-querying a date multiple times.

PR #82, opened 2024-11-08
bugfix/handle single verse readings and prevent infinite loop

andylitalo:
**Bugfixes**
- Single-verse readings were failing to be parsed because the regex expected a `"-"`: now, make that an optional group
- To prevent infinite loop, advance through html content immediately
- add try-except around group parsing just in case

**Testing**
Get the readings for 2024-11-25 and check that they are parsed properly (this day includes each type of reading: single-chapter, single-verse, and spanning chapters)

You can do this by running the server locally
```
python manage.py runserver
```
And retrieving the endpoint at `http://localhost:8000/hub/readings/?date=2024-11-25`


mattash (approved):
Test, works great. How did you find this bug?

PR #83, opened 2024-11-08
[bugfix] raise ValidationError if user tries to register with existing email

andylitalo:
**Testing**

Try to register a new user with an existing email: expect a validation error message
Try to register a new user with a non-existing email: expect successful registration


mattash (approved):
Works as expected.

PR #87, opened 2024-11-12
Feature/log in with email

andylitalo:
**Feature**

Custom backend to log in with email instead of username.

Essentially, it treats "username" field passed by login form as an email.

**Test**
Try logging in with email (expect success), then username (expect failure) for a user.


andylitalo:
Cut from wrong branch to merge into master so it contains extraneous commits. Created new branch with cleaner cut--will make PR.

PR #88, opened 2024-11-12
[feature] log in with email instead of username

andylitalo:
**Feature**

Custom email backend to log in users with their email instead of their username.

Essentially treats username as an email.

**Test**
Try to log in as a user with their email. Log out. Then try to log in with their username. Both should work.


mattash (approved):
Works as expected!

PR #98, opened 2024-11-23
Bugfix/notifications task fix

mattash:
Previous edits to the send_push_notification function didn't work when tested with live data. The function was refactored and tested.

This pull request includes updates to the notification system to improve the handling of push notifications. 

Improvements to notification handling:

* [`notifications/tasks.py`](diffhunk://#diff-e3ebb51888cfce5bc4b2fc128b7bd14f7cab4a7139b7a8bb5129dc419fe6b239R14-R16): Added a TODO comment to address the need for updating tasks to send notifications to all churches when there are multiple churches.

Code simplification and optimization:

* [`notifications/utils.py`](diffhunk://#diff-c300da093d8bb0cb0420b1781d72d9c37f9593ce7dba9b58660e8dd4df98f19dL42-R53): Refactored the `send_push_notification` function to use a dictionary for filtering based on user preferences, reducing repetitive code and improving maintainability.

### DIFF
--- a/notifications/utils.py
+++ b/notifications/utils.py
@@ -39,29 +39,22 @@ def send_push_notification(message, data=None, users=None, notification_type=Non
         if users:
             tokens_queryset = tokens_queryset.filter(user__in=users)
 
-        # Filter based on user preferences if notification_type is provided
-        if notification_type == 'upcoming_fast':
-            tokens_queryset = tokens_queryset.filter(
-                user__profile__receive_upcoming_fast_push_notifications=True
-            )
-        elif notification_type == 'ongoing_fast':
-            tokens_queryset = tokens_queryset.filter(
-                user__profile__receive_ongoing_fast_push_notifications=True
-            )
-        elif notification_type == 'daily_fast':
-            tokens_queryset = tokens_queryset.filter(
-                user__profile__receive_daily_fast_push_notifications=True
-            )
-        elif notification_type == 'weekly_fast':
-            tokens_queryset = tokens_queryset.filter(
-                user__profile__include_weekly_fasts_in_notifications=True
-            )
+        if notification_type:
+            notification_type_filters = {
+                'upcoming_fast': 'receive_upcoming_fast_push_notifications',
+                'ongoing_fast': 'receive_ongoing_fast_push_notifications',
+                'daily_fast': 'receive_daily_fast_push_notifications',
+            }
 
-        if notification_type != 'weekly_fast' and not data.get('weekly_fast', False):
-            tokens_queryset = tokens_queryset.exclude(
-                Q(user__profile__fasts__name__icontains='friday') |
-                Q(user__profile__fasts__name__icontains='wednesday')
-            )
+            if notification_type in notification_type_filters:
+                filter_field = notification_type_filters[notification_type]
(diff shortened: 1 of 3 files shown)


notifications/utils.py
```diff
@@ -39,29 +39,22 @@ def send_push_notification(message, data=None, users=None, notification_type=Non
...
-        elif notification_type == 'daily_fast':
-            tokens_queryset = tokens_queryset.filter(
-                user__profile__receive_daily_fast_push_notifications=True
-            )
-        elif notification_type == 'weekly_fast':
-            tokens_queryset = tokens_queryset.filter(
-                user__profile__include_weekly_fasts_in_notifications=True
-            )
+        if notification_type:
+            notification_type_filters = {
```
andylitalo, line 43: I think we should break this out into a constant so it's easier to find for future updates

notifications/utils.py
```diff
@@ -39,29 +39,22 @@ def send_push_notification(message, data=None, users=None, notification_type=Non
...
-        if notification_type != 'weekly_fast' and not data.get('weekly_fast', False):
-            tokens_queryset = tokens_queryset.exclude(
-                Q(user__profile__fasts__name__icontains='friday') |
-                Q(user__profile__fasts__name__icontains='wednesday')
-            )
+            if notification_type in notification_type_filters:
+                filter_field = notification_type_filters[notification_type]
+                tokens_queryset = tokens_queryset.filter(**{f'user__profile__{filter_field}': True})
+
+            if not tokens_queryset.filter(user__profile__include_weekly_fasts_in_notifications=True).exists():
```
andylitalo, line 53: I think this should also evaluate to true if the notification type is not weekly fast?
```
if notification_type != "weekly_fast" or not tokens_queryset.filter(...):
```

mattash: so, we don't need a 'weekly_fast' notification type for this function. the weekly_fast setting in user profile (`user__profile__include_weekly_fasts_in_notifications`) is going to filter out weekly fasts if it isn't True. But we would run send_push_notification(message, data, user, 'weekly_fast'). In other `notification_type != "weekly_fast"` isn't an actual case.

notifications/utils.py
mattash: Resolved in [c4a8fc7](https://github.com/surp-hovhannes/bahk/pull/98/commits/c4a8fc79259e94dd4231e4ac67dfcf09e85444a9)

notifications/utils.py
andylitalo: Gotcha, I see now that `weekly_fast` isn't every passed as a `notification_type`--could remove it from one of the options listed in the docstring for that arg?

andylitalo: I'm still missing something. The way I understand the logic (may be wrong) is that if there is no token for any of the users that have requested weekly fast notifications, we should exclude all tokens for users that have signed up for a weekly fast. 

It seems then that if there is only one user and they haven't requested weekly fast notifications but have signed up for a weekly fast, all of their tokens will be removed from the queryset in this method? Am I missing something?

mattash:
I think we've got ourselves a winner! I tested this with all scenarios including weekly fasts.

andylitalo (approved):
LGTM! Thanks for workshopping through it

PR #107, opened 2024-12-05
Issue102/longer jwt token expiration lifetime

andylitalo:
Makes lifetime of simple JWT token essentially eternal (100 years) so users only have to log in once.

Tested by changing lifetime to 30 seconds and showing that the refresh token is valid at first, then expires after 30 seconds when passed to the refresh API endpoint.

### DIFF
--- a/bahk/settings.py
+++ b/bahk/settings.py
@@ -169,7 +169,7 @@ REST_FRAMEWORK = {
 
 # extend lifetime of JWT refresh tokens
 SIMPLE_JWT = {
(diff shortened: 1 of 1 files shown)


mattash:
Is there an advantage to not setting "ACCESS_TOKEN_LIFETIME" too?

andylitalo:
No, but I don't think it's necessary since you can obtain a new access token with the refresh token. If you want to extend it, we can--it might save some API calls to refresh the token.

PR #112, opened 2024-12-27
[bugfix] Handles creating fast with dates if dates exist

andylitalo:
**Bug**

Previously, creating fast with dates was treating days as having a many-to-many relationship with fasts, which was causing uncaught errors and leading to cryptic Server 500 errors. These expressions should have been updated when the `Day` model was updated to have foreign key relationships with `Fast`.

**Fix**

Check for overlapping days of fast using foreign key relationship
Create link between day and fast upon creation of day by setting the fast as its foreign key.

**Testing**

1. Create a fast with dates (go to admin > `Fasts` > `CREATE FAST WITH DATES` in the upper right)
2. Create another fast with dates with a new name but same church and overlapping at least 1 day with the fast you just created. You should get a validation error indicating that there is an overlap with an existing fast from that church. Check that `Fasts` page to make sure that the fast was not created without the dates (this was happening before).
3. Adjust the dates so they do not overlap with a fast from your current church, but they do overlap with a fast from another church. The fast should be created successfully.

### DIFF
--- a/hub/admin.py
+++ b/hub/admin.py
@@ -135,11 +135,11 @@ class FastAdmin(admin.ModelAdmin):
                 fast = form.save()
                 data = form.cleaned_data
 
-                # days
+                # create days for fast
                 dates = [data["first_day"] + datetime.timedelta(days=num_days) 
                          for num_days in range(data["length_of_fast"])]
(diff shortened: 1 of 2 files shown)


mattash (approved):
Lgtm!

Issue #113, opened 2024-12-28
Error retrieving readings from St. James Epistle


mattash:
occurred when attempting to retrieve readings for December 28,2024

[Safari - Dec 28, 2024 at 12:56 PM.pdf](https://github.com/user-attachments/files/18267679/Safari.-.Dec.28.2024.at.12.56.PM.pdf)

PR #114, opened 2024-12-29
Bugfix/add missing link for st james on catena

andylitalo:
**Bug**

On December 28, 2024, readings could not be fetched because of an unforeseen name for the book of James, which caused creation of the Catena URL to fail, failing the fetching of the rest of the readings.

**Fix**

This name has been added to the list and a backup has been added to avoid this in the future.

Tested locally by andylitalo.

Resolves issue https://github.com/surp-hovhannes/bahk/issues/113


andylitalo:
Resolves Issue 113: https://github.com/surp-hovhannes/bahk/issues/113

mattash (approved):
Good idea to provide a fallback url

PR #115, opened 2024-12-30
[bugfix] ensure seed church pk starts with 1

andylitalo:
**Bug**

When seeding database after having already created a database, might skip `pk=1` when creating churches. This is necessary to fetch churches when not logged in (defaults to fetching church with `pk=1`).

**Fix**

Hard-coded `pk`s for seed churches (1,2,3).

Tested.


mattash (approved):
This makes sense.

PR #116, opened 2024-12-30
[bugfix] allow no tz specification (defaults to UTC)

andylitalo:
**Bug**

If no timezone specified in query params (`?tz=<timezone>`), `self.context['tz']` called in the `FastSerializer` raises an error.

**Fix**

Use the safe key retrieval, `self.context.get('tz')`. If not provided, the `timezone` library assumes UTC.

Tested locally by @andylitalo


mattash (approved):
This is a good find. 👍

PR #117, opened 2024-12-30
Feature/email for username backend

andylitalo:
**Feature**

Obsoletes use of "username", replacing with email where it cannot be completely removed (such as in the `User` class).

Should only be deployed with [companion PR for frontend](https://github.com/surp-hovhannes/bahk_react_native/pull/54).

To test,
```
python -m pip install -r requirements.txt
python manage.py seed
python manage.py runserver <local-machine-ip-address>:8000
```

You may need to change `settings.py` so `ALLOWED_HOSTS` includes your machine's IP address.

Then, in the `bahk_react_native` repo, with the companion branch `feature/email-for-username-frontend` active, change `API_HOST = 'http://<ip-address>:8000'` in `constant/constants.js` and run
```
npx expo start
```
Check that you can register a new user and login as an existing user (e.g., `email: user1a@email.com`,  `password: default123`).

### DIFF
--- a/bahk/settings.py
+++ b/bahk/settings.py
@@ -172,8 +172,8 @@ SIMPLE_JWT = {
     'REFRESH_TOKEN_LIFETIME': datetime.timedelta(days=100*365)  # set to expire in 100 years (~forever)
 }
 
-# path to backend for authenticating users by email or username (instead of only username)
-AUTHENTICATION_BACKENDS = ["hub.auth.EmailOrUsernameBackend"]
+# path to backend for authenticating users by email
+AUTHENTICATION_BACKENDS = ["hub.auth.EmailBackend"]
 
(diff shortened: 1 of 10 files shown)


mattash:
I will test this more thoroughly, but a general question about approach. Do we want to provide the user a "name" field, even if it's not a username? Right now user exposure is limited, at most inclusion in the participants list for a given fast. But the future will more likely bring more user generated content being displayed. Should users have a handle that is displayable, and not their email address?

andylitalo:
Good point--I was thinking about how to include that in this PR. How about allowing the user to provide a first name and a last name/initial? I'll draft a separate PR for this that can be merged into this one.

andylitalo:
Separate PR submitted with single "name" field.

mattash (approved):
This works as expected. Great job!

PR #130, opened 2025-02-04
Add section for generating static content in README

evanugarte:
I had to run this command for the html to show up correctly at http://localhost:8000/hub/ and http://localhost:8000/hub/web

If this is a one-off issue, we don't have to add this and we can close the pr

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,12 @@ python manage.py createsuperuser
 ```
 Note the username, email, and password that you enter.
 
+### Running the website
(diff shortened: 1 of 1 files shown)


andylitalo:
No harm in adding it! You need to run this if you start from an empty database (which is why Evan saw it on his first setup while Der Mesrop does not).

andylitalo (approved):
Helpful note to add!

PR #131, opened 2025-02-04
Specify emails of default accounts

evanugarte:
We can reword/reformat the changed lines

since the login page requires an email over a username, I thought we could specify the email of the default accounts. I had to guess the email of each user when I first ran the app

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,14 @@ First, populate the database with the seed data by running
 python manage.py seed
 ```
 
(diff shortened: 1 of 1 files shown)


mattash (approved):
Makes sense to me!

andylitalo (approved):
My bad! I've forgotten to update the README in the latest PRs. Thanks for documenting.

PR #132, opened 2025-02-05
Some enhancements for Admin interface

mattash:
After using the admin interface for a while now in production, these are a few enhancements that I think would be nice to have. Let me know if you guys think of anything else.


andylitalo (approved):
This is great! Interface is much cleaner.

PR #128, opened 2025-02-03
Add thumbnail caching mechanism for models with images

mattash:
- Introduced cached_thumbnail_url and cached_thumbnail_updated fields for Fast, Profile, Video, and Article models
- Created ThumbnailCacheMixin to handle thumbnail URL caching
- Added management command to update thumbnail caches
- Optimized serializers to use cached thumbnail URLs
- Improved performance by reducing S3 calls for image thumbnails
- Updated runtime to Python 3.11.7 and added new dependencies
- Configured Redis caching for Django and Celery

Primary thing to test: adds images to affected models, verify that thumbnails are generated and saved to thumbnail cache fields.

This pull request introduces several significant changes to the `bahk` project, focusing on adding caching mechanisms for thumbnails, improving middleware configuration, and enhancing logging. Here are the most important changes:

### Middleware and Configuration Enhancements:
* Added `CorsMiddleware` to the `MIDDLEWARE` list in `bahk/settings.py` to handle Cross-Origin Resource Sharing (CORS).
* Introduced a helper function and configuration for Redis URL and cache settings, including SSL support for Redis.
* Updated static files configuration, changing `STATIC_URL` to `/static/`.
* Enhanced Celery configuration to include SSL settings when using Redis over SSL.
* Improved environment logging with colored output to distinguish between production and development environments.

### Thumbnail Caching:
* Added a new management command `update_thumbnail_cache` to update cached thumbnail URLs for models with images.
* Introduced migrations to add `cached_thumbnail_updated` and `cached_thumbnail_url` fields to the `Fast` and `Profile` models. [[1]](diffhunk://#diff-0c928ba0a6947e6e87e7d1c863a874964d64c5fda02187207e6d46751eaad494R1-R23) [[2]](diffhunk://#diff-4ad620a376a2644f31c56873358432372bda9d44efc4aee77420e3b09cceca2bR1-R23)
* Implemented `ThumbnailCacheMixin` to handle thumbnail caching for models with `ImageSpecField` thumbnails.
* Updated `Fast` and `Profile` models to use the new caching mechanism, including tracking changes to image fields. [[1]](diffhunk://#diff-2e0cf63c3374e524fb024d4540dac9f4f555a8db27d993f35e84595c31495d0aL50-R108) [[2]](diffhunk://#diff-2e0cf63c3374e524fb024d4540dac9f4f555a8db27d993f35e84595c31495d0aR143-R196)

### Serializer Enhancements:
[...]

PR #133, opened 2025-02-07
fix: timezone handling in FastDetailView

mattash:
PR #128 introduced a bug, whereby FastDetail caching required a timezone be defined, and the existing fallback to UTC wasn't generating an object with .zone. Requiring a timezone breaks the app because timezone is only explicitly queried to check for the fast of the current day. This change restored expected functionality.  

- Replace timezone.utc with pytz.UTC to ensure .zone attribute availability
- Add timezone string sanitization to handle malformed query parameters
- Maintain cache key generation functionality which depends on timezone.zone attribute

### DIFF
--- a/hub/views/mixins.py
+++ b/hub/views/mixins.py
@@ -40,4 +40,4 @@ class TimezoneMixin:
         tz_str = self.request.query_params.get('tz')
         if tz_str:
             return pytz.timezone(tz_str)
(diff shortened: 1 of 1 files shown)


andylitalo:
LGTM--how should we test?

andylitalo (approved):

PR #135, opened 2025-02-22
Draft for devotional model and view

andylitalo:
**Features**

- `Devotional` model: stores a video with title and description
- `DevotionalSet`: holds an ordered set of devotionals (for fast-based devotionals)
- `DevotionalSerializer`: serializes `Devotional` object (no serializer for `DevotionalSet` yet)
- `DevotionalByDate` view: returns serialized devotional data by date and church

Devotional creation and `DevotionalByDate` endpoint have been tested--try it out yourself and suggest improvements! This should be developed to match the desired data in the frontend.


mattash (changes requested):
Awesome! Just glancing at the code, one thing that comes to mind that I know we'll need is an endpoint that returns devotionals by fast. That way we can show all the devotionals associated with a fast on the fast details page.

this component will display a list of devotionals using the fastId property: https://github.com/surp-hovhannes/bahk_react_native/blob/23cb6fc0f91fbd36a2b054b73b0ba9f8e114734c/components/DevotionalsList.tsx

andylitalo:
Added! I added a full listview of devotionals by church for completeness. We might end up just grouping devotionals by fast and not needed to create additional DevotionalSets, but I'm leaving that model in there just in case, although I can remove if you prefer.

andylitalo:
Merged master and merged migrations into one

PR #139, opened 2025-02-27
Recipe model, serializer, view, endpoint

andylitalo:
Someone requested recipes so I took a crack at adding them to the backend. I based the `Recipe` model on the `Article` model and envision the first draft showing them below the articles on the Learn tab. I didn't have time to test it, so very much a draft, but wanted to put it out there in case I don't get a chance to work on it for a while.

EXPECT BUGS. UNTESTED.


andylitalo:
Got this working with the accompanying [frontend PR](https://github.com/surp-hovhannes/bahk_react_native/pull/110)
Just couldn't get the photo to load but might have to do with storing it locally (and not using redis...)
image

mattash (approved):
I didn't notice any bugs. This is functionally a good v1 for me. I wonder if the recipe model could use a description or something. Not instructions or ingredients, but something more narrative.

andylitalo:
Added description section!

PR #140, opened 2025-03-03
Enhancement/multipart-upload-for-videos

mattash:
using a django admin plugin to handle large file uploads directly to AWS s3.


andylitalo (approved):
Added missing django-s3direct package and merged video migrations into one file. All good!

PR #168, opened 2025-04-22
[feature] allow admin to specify email addresses to send promo email to

andylitalo:
**Testing**

I've tested that the new field `specific_emails` is properly validated, but I haven't been able to send emails from my local server. I'm not sure if it's an issue with my local environment or with the `send_now()` admin action not working properly.


andylitalo:
Set up celery server locally--successfully received email to manually selected recipients!

mattash (changes requested):
This works as expected, however our current need would be better served by being able to indicate who shouldn't receive e-mails. In the first round of emails that went out yesterday, roughly 120 emails were sent, so we need to be able to indicate that we don't want those email addresses to be sent. I'm thinking that a checklist of all emails would be flexible enough to fix our current issue and be useful in the future.

mattash:
I implemented the change I requested. @andylitalo can you cleanup the migrations?

andylitalo:
Merged migrations--works!

mattash (approved):
Yay for us. Lets go.

PR #172, opened 2025-04-26
[feature] integrate OpenAI context generation for readings

mattash:
This pull request introduces an AI-powered contextual introduction feature for Bible readings, leveraging OpenAI's API. It includes changes to models, tasks, views, and tests to support the generation, storage, and feedback for AI-generated context. Below are the most important changes grouped by theme:

### AI Context Generation

* Added `OPENAI_API_KEY` and `READING_CONTEXT_REGENERATION_THRESHOLD` to settings for OpenAI integration. (`bahk/settings.py`)
* Created `generate_context` function in `hub/services/openai_service.py` to interact with OpenAI's API and generate contextual introductions for Bible passages. (`hub/services/openai_service.py`)

### Database and Model Updates

* Added fields `context`, `context_last_generated`, `context_thumbs_up`, and `context_thumbs_down` to the `Reading` model for storing AI-generated context and user feedback. (`hub/migrations/0027_reading_context_reading_context_last_generated_and_more.py`, `hub/models.py`) [[1]](diffhunk://#diff-5e6df8227a67e9401379e4fe6a8261d14678514ae26e068519891454623a8ae6R1-R33) [[2]](diffhunk://#diff-2e0cf63c3374e524fb024d4540dac9f4f555a8db27d993f35e84595c31495d0aR329-R332)

### Task Management

* Introduced a new Celery task `generate_reading_context_task` to asynchronously generate and save AI-generated context for readings. (`hub/tasks/openai_tasks.py`, `hub/tasks.py`, `hub/tasks/__init__.py`) [[1]](diffhunk://#diff-43f590ea13178e234393a4ba31f70fc53bc3364eb988be2bf5d92477bd333ef7R1-R44) [[2]](diffhunk://#diff-5000b4b9d0f45f97fcba5492050eb835b48962b9d71a8ad1418f3d719cdab8aeR12) [[3]](diffhunk://#diff-828a7539c6eebfca0ad8386f9071ea145bf15770098a1cde0d6332751ba7bbbaR9-R18)

### API and Feedback Mechanism

* Enhanced the `GetDailyReadingsForDate` API to include context fields and trigger context generation if missing. (`hub/views/readings.py`)
* Added a new `ReadingContextFeedbackView` API endpoint to handle user feedback (thumbs up/down) for AI-generated context and trigger regeneration when down-votes exceed the threshold. (`hub/views/readings.py`, `hub/urls.py`) [[1]](diffhunk://#diff-9538672adba91e6696ee5c0a455f0d5fe49f8e8e93978a92d96495d4780fd356R11-R15) [[2]](diffhunk://#diff-843c3379d21c14761bbcf2ccee37b6ca26a4de0d9141c97fe65c038ca8b9debbR101)

### Testing

* Added unit tests for the context generation task, API integration, and feedback mechanism to ensure functionality and reliability. (`hub/tests/test_reading_context.py`)


hub/models.py
```diff
@@ -326,6 +326,10 @@ class Reading(models.Model):
     start_verse = models.IntegerField(verbose_name="Start Verse")
     end_chapter = models.IntegerField(verbose_name="End Chapter", help_text="May be same as start chapter")
     end_verse = models.IntegerField(verbose_name="End Verse", help_text="May be same as end verse")
+    context = models.TextField(null=True, blank=True, help_text="AI-generated contextual introduction")
```
andylitalo, line 329: What do you think about creating a model for `AIContext` with these fields? It can have a `ForeignKeyField` that points to a `Reading`. Additionally, each `Reading` could have a `OneToOneField` that points to the preferred `AIContext` that will be sent over the API call. This way we can keep a record of past `AIContext`s and their voting history in case we want go back to a previous one or use them to improve context generation in the future.

hub/services/openai_service.py
```diff
@@ -0,0 +1,52 @@
...
+        return None
+
+    prompt = PROMPT_TEMPLATE.format(passage=passage_reference)
+    try:
+        response = openai.chat.completions.create(
+            model="gpt-4.1-mini",
+            messages=[
+                {
+                    "role": "system",
+                    "content": "You are a helpful assistant providing biblical context with historical and canonical background only."
```
andylitalo, line 40: What about adding the following?
"You believe completely in the inspiration of the Holy Bible by the Holy Spirit and that Jesus Christ is the Son of God, Our Lord and Our Savior."

hub/services/openai_service.py
```diff
@@ -0,0 +1,52 @@
...
+from django.conf import settings
+
+logger = logging.getLogger(__name__)
+
+
+# Ensure the API key is configured
+openai.api_key = getattr(settings, "OPENAI_API_KEY", None)
+
+
+PROMPT_TEMPLATE = (
```
andylitalo, line 14: I think we will want different prompts for different books of the Bible (specifically Psalms and Proverbs come to mind). What do you think of assigning the prompt based on the book?

andylitalo (commented):
A few ideas--I can prototype them on a separate branch shortly.

hub/services/openai_service.py
andylitalo: Never mind, the prompt I'm working on works well with wisdom literature.

hub/services/openai_service.py
```diff
@@ -0,0 +1,52 @@
...
+import openai
+from django.conf import settings
+
+logger = logging.getLogger(__name__)
+
+
+# Ensure the API key is configured
+openai.api_key = getattr(settings, "OPENAI_API_KEY", None)
+
+
```
andylitalo, line 13: The prompt template should be stored in the db instead of hard-coded so we can improve over time. One idea is to create a `Prompt` model. There will be one `active` `Prompt` that is used for generating new prompts, which is by default the latest created. Each `AIContext` object will point to its `Prompt` via a `ForeignKeyField`. `AIContext` generated by passed `Prompt`s can still be saved and used as the primary context for a reading, but all new `AIContext`s will be generated with the active `Prompt`.

hub/models.py
mattash: I would recommend making the model name, more descriptive in terms of it's association something like ReadingContext. But this is a good idea. I also noticed that it would be hard to iterate if we weren't keeping track of the contexts that we didn't like.

hub/services/openai_service.py
mattash: Good idea. I'm interested in exploring the tone. I don't want the scripture contextualized from an atheistic point of view. But I'm also not feeling good about the AI acting like a Christian. 🤔

I think we need to dog food it and just get a feeling for what is working for us in practice.

hub/services/openai_service.py
mattash: Thats smart. I also recommend that the prompt have a 'model' attribute or 'openai_model'. We can play around with them to how it affects the prompts. And we can step down the model if prices go up.  It would be an enum and for now the options are:
```
'gpt-4.1-mini'
'gpt-4.1-nano'
'gpt-4o-mini' 
'gpt-4.5-preview'

andylitalo: Ah good idea. We can wrap all the params for the API call into what I was calling the `Prompt` model (maybe call it `OpenAIParams` since we're just using `openai`'s API?), which could include `max_tokens`, `temperature`, and `role`.

While I'm thinking about this, if there's any chance we'd want to switch from OpenAI to another LLM provider, we could use the `langchain` package instead of `openai`, which supports [many chat models](https://python.langchain.com/docs/integrations/chat/)

mattash: I'm not married to openai. I already have the developer access setup so it was the most straightforward to implement.

PR #176, opened 2025-05-12
Mattash/enhancement-llmprompt-admin-actions

mattash:
adds actions to duplicate an existing prompt, useful for iteration, and make a selected prompt active from the index of prompts. Also hads helpful fields to the list of prompts.


andylitalo (approved):
LGMT!

PR #177, opened 2025-05-21
Mattash/chatgpt fine tuning

mattash:
This pull request introduces functionality for comparing LLM prompts for a selected reading, alongside updates to improve prompt management and enhance the OpenAI service. The most significant changes include adding a new admin action for comparing prompts, updating the LLM prompt model with additional choices, modifying the OpenAI service to refine prompt handling, and creating a new admin view and template for prompt comparison.

### New Feature: Compare LLM Prompts
* Added a new admin action `compare_prompts` in the `ReadingAdmin` class to allow redirection to a page for comparing LLM prompts for a selected reading. The action validates that exactly one reading is selected before proceeding. (`hub/admin.py`)
* Created a new view `compare_reading_prompts` to handle the comparison of prompts, which generates and displays outputs for selected prompts. (`hub/views/admin.py`)
* Added a new URL pattern for the `compare_reading_prompts` view. (`hub/urls.py`)
* Designed a new HTML template `compare_reading_prompts.html` for rendering the comparison interface, including prompt selection and output display. (`hub/templates/admin/compare_reading_prompts.html`)

### Enhancements to LLM Prompt Management
* Updated the `LLMPrompt` model to include new options in `MODEL_CHOICES`, such as `gpt-4.1` and `gpt-o4-mini`. (`hub/models.py`)

### Improvements to OpenAI Service
* Refactored the `generate_context` function to improve prompt construction and modify parameters for generating OpenAI completions. Changes include combining the `role` and `prompt` fields, adjusting the user prompt, and updating parameters like `max_tokens` and `temperature`. (`hub/services/openai_service.py`)


hub/services/llm_service.py
```diff
@@ -0,0 +1,101 @@
...
+        """Generate context for a reading using the LLM service."""
+        pass
+
+class AnthropicService(LLMService):
+    """Service for Anthropic's Claude API."""
+    
+    def __init__(self):
+        self.client = anthropic.Anthropic(api_key=settings.ANTHROPIC_API_KEY)
+    
+    def generate_context(self, reading: Reading, llm_prompt: Optional[LLMPrompt] = None) -> Optional[str]:
```
andylitalo, line 26: This code appears to be duplicated in `hub/services/anthropic_service.py`. Should it live here or should it import from that file?

mattash:
Good point. Actually the anthropic service and OpenAI service files are deprecated for the I’ll service file. I think that does all the work now.

hub/tasks/openai_tasks.py
andylitalo, line 1: Would it be too disruptive to rename to `llm_tasks.py`?

hub/services/llm_service.py
```diff
@@ -0,0 +1,101 @@
...
+                store=True
+            )
+            if response.choices:
+                return response.choices[0].message.content.strip()
+            logger.error("OpenAI response contained no choices")
+        except Exception as exc:
+            logger.exception("OpenAI API call failed: %s", exc)
+        return None
+
+def get_llm_service(model_name: str) -> LLMService:
```
andylitalo, line 94: Given that we'll always generate context using an `LLMPrompt` instance, what do you think about moving the method into the `LLMPrompt` model class and getting the `model_name` from its `model` field? The call signature would then be

```
service = llm_prompt.get_llm_service()
context = service.generate_context(...)
```

and spare the import for `get_llm_service`

andylitalo (commented):
A few suggestions but this looks really helpful!

hub/services/llm_service.py
mattash: you got it

mattash:
> A few suggestions but this looks really helpful!

Thanks! Those were great. Take a look and let me know if there is anything else.

PR #178, opened 2025-05-31
feature: draft for internationalizing learning resources

andylitalo:
## Motivation

I thought we could start building out infrastructure for translations. The learning resources was the smallest chunk to bite off.

Makes two major design choices:
1. Store translations as a *separate* model from the original (`Article` vs. `ArticleTranslation`). This makes English preferred, so the logic for handling the language is a little more complex, but simplifies the migration by not changing the existing structure
2. If a translation does not exist, no content is provided. We could also provide English as a default to so untranslated content is still available, but I thought this might look like a bug.

Tested the `ListView` APIs and played around with the admin page, but some more testing wouldn't hurt.


mattash (commented):
Before we go to far down this path, I did a little searching and it seems like there is a widely adopted Django module that facilitates this multi-lingual translation according to be best practices: https://github.com/deschler/django-modeltranslation. Do you to take a look and compare pros and cons of each approach?

andylitalo:
Stale branch--need to research other options.

PR #185, opened 2025-06-05
fix: add correct name of the book of Jude

andylitalo:
**Bug**

Mapped the book named "Jude" to the URL segment, but book was actually named "St Jude's General Epistle." This was not noticed before because this is the first day of the year when we read from Jude.

**Fix**

Added the correct name as the key to map to URL segment for the catena url.

**Testing**

My local environment is buggy, so I haven't tested yet.


andylitalo:
Added correct name for John 3 too

PR #196, opened 2025-06-20
[bugfix] Don't send reminder emails if promo email assigned to the fast

andylitalo:
**Bug**

Users would receive 3 automated reminder emails for an upcoming fast in addition to promotional emails sent about the devotional series for that fast.

**Fix**

Don't send reminder emails for fasts that already have a promotional email assigned to it (I added a foreign key to the `PromoEmail` model to optionally assign it to a `Fast`).

**Testing**

Added a test that no reminder emails are sent if there is a promotional email assigned to the fast.


mattash (approved):
Good solution! I added some help text for the fast selection field.

PR #295, opened 2026-02-12
Feature/fums and compliance

andylitalo:
**Frontend changes for FUMS merged**

# Overview

## What we needed to do

API.Bible's Terms of Use require implementers to participate in their **Fair Use Management System (FUMS)** — an anonymized tracking system that lets rights holders and publishers monitor scripture usage. Before exposing Bible text to end users, we must:

1. Request FUMS v3 tokens from API.Bible with every passage request
2. Store those tokens so the frontend can report them when scripture is displayed
3. Serve the tokens to the frontend via the readings API

Without FUMS compliance, our API.Bible access could be revoked.

## Why we needed to do it

Our existing integration fetched scripture text and stored it, but never captured or forwarded FUMS tokens. This put us out of compliance with API.Bible's terms. The Terms of Use (Sections 3, 5, 8, 10) require FUMS tracking for any application that displays scripture to users.

### FUMS token capture (`hub/services/bible_api_service.py`)
- Added `fums-version=3` to all API.Bible request parameters
- Extract `fumsToken` from the `meta` field of every API.Bible response
- Return it alongside the passage content

### Database storage (`hub/models.py`, migration `0046`)
- Added `fums_token` TextField to the `Reading` model
- Created migration `0046_add_fums_token_to_reading`

### Celery task updates (`hub/tasks/bible_api_tasks.py`)
- `_fetch_and_update_passage()` now saves the FUMS token when updating readings
- Deduplication logic copies `fums_token` from existing readings with the same passage

### API response (`hub/views/readings.py`)
- The readings endpoint now includes `fumsToken` in each reading's JSON response

### Admin display (`hub/admin.py`)
- Added a `has_fums_token` boolean column (checkmark) instead of showing the raw token
- Added `fums_token` to `readonly_fields` and excluded it from the edit form
- Added a description explaining FUMS compliance in the "Bible Text" fieldset

### Documentation (`README.md`)
- Replaced the placeholder FUMS note with comprehensive compliance documentation covering FUMS, DRM, copyright, security, non-commercial use, and logo policy

## Testing

- [x] **Migration**: Apply migration `0046` to local SQLite database; verified the column was created and the 500 error (caused by missing column `hub_reading.fums_token`) was resolved.
- [x] **Token capture**: After deploying the change, verify via Django admin that new readings had FUMS tokens. For existing readings that were fetched before the `fums-version=3` parameter was added, we cleared `text_fetched_at` and triggered `refresh_all_reading_texts_task` to re-fetch them with tokens.
- [x] **API response**: Confirm the frontend receives `fumsToken` in the readings API JSON response.
- [x]**Admin UX**: Verified the boolean checkmark displays correctly in the admin list view and that the FUMS description appears in the fieldset.

### DIFF
--- a/events/admin.py
+++ b/events/admin.py
@@ -215,6 +215,7 @@ class EventAdmin(admin.ModelAdmin):
             path('analytics/new-users/', self.admin_site.admin_view(self.new_users_data), name='events_new_users_data'),
             path('analytics/devotional-views/', self.admin_site.admin_view(self.devotional_views_data), name='events_devotional_views_data'),
             path('analytics/checklist-usage/', self.admin_site.admin_view(self.checklist_usage_data), name='events_checklist_usage_data'),
+            path('analytics/checklist-items/', self.admin_site.admin_view(self.checklist_items_data), name='events_checklist_items_data'),
             path('analytics/prayer-views/', self.admin_site.admin_view(self.prayer_views_data), name='events_prayer_views_data'),
             path('export_csv/', self.admin_site.admin_view(self.export_csv), name='events_export_csv'),
         ]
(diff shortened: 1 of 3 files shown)


bahk/settings.py
```diff
@@ -473,7 +473,8 @@ def is_redis_ssl(url):
     print("\033[93m" + "="*50 + "\033[0m")  # Reset color
 
 # handling CORS headers
-CORS_ORIGIN_ALLOW_ALL = config('CORS_ORIGIN_ALLOW_ALL', default=False, cast=bool)
```
andylitalo: I was having issues with CORS permissions on Django 4.7. `CORS_ALLOW_ALL_ORIGINS` is apparently the new env var, but I'm leaving the old one for compatibility.

andylitalo:
> @andylitalo I ran a branch review and this comment came back. I don't know how this FUMS system works, but is this a valid issue?
> 
> 1. Stale FUMS tokens on copied readings (bible_api_tasks.py:105-111): When a reading is copied from an existing one (deduplication), it also copies the fums_token. FUMS tokens are typically single-use or session-bound - is it valid to reuse the same token across multiple readings? If each "view" needs a unique FUMS report, the frontend may need to handle this, but it's worth confirming with the API.Bible docs.

Ah good point: we should not copy the token. I didn't make that specification explicit and will update accordingly.

PR #310, opened 2026-02-21
Add start_day_number to Fast to support Day 0 fasts

mattash:
## Summary

- Adds a `start_day_number` field (default=1) to the `Fast` model
- Updates `get_current_day_number` in the serializer to apply `start_day_number - 1` as an offset on both the annotation and fallback computation paths
- Fixes the off-by-one error in `current_day_number` for fasts that begin with a "Day 0" devotional (e.g., Fast 112 / Great Lent 2026)

**Backward compatible:** all existing fasts default to `start_day_number=1`, which produces identical behavior to before.

To fix an affected fast, set `start_day_number=0` via the admin.

## Test plan

- [x] Existing hub unit tests pass
- [ ] Verify fast 112 returns `current_day_number=6` after setting `start_day_number=0` in admin
- [ ] Confirm other fasts are unaffected

---

> [!NOTE]
> **Low Risk**
> Low risk: simple additive model field plus a small arithmetic adjustment in a serializer; existing fasts keep identical behavior due to the default value.
> 
> **Overview**
> Adds a `start_day_number` field to `Fast` (via migration) so a fast can start at day 0 (or other non-1 numbering) while defaulting to 1 for existing data.
> 
> Updates `FastSerializer.get_current_day_number` to apply a `(start_day_number - 1)` offset to both the annotated `current_day_count` path and the fallback computed count, fixing off-by-one day numbering for “Day 0” fasts. The migration also updates `Devotional.order` field metadata/help text.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 11ebda5b50918f83095d598a26998bdf3fe7dad3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/hub/models.py
+++ b/hub/models.py
@@ -69,9 +69,9 @@ class Fast(models.Model):
         blank=True,
         help_text="Attribution or author of the culmination feast message"
     )
-    start_day_number = models.PositiveIntegerField(
-        default=1,
-        help_text="The day number assigned to the first day of the fast (e.g., 0 if the fast has a 'Day 0')",
+    has_day_zero = models.BooleanField(
+        default=False,
+        help_text="If True, the fast starts counting from Day 0 instead of Day 1",
     )
 
     # auto-saved to be the year of the first day of the fast
--- a/hub/serializers.py
+++ b/hub/serializers.py
@@ -270,13 +270,16 @@ class FastSerializer(serializers.ModelSerializer, ThumbnailCacheMixin):
         return None
     
     def get_total_number_of_days(self, obj):
-        """Use annotated value if available"""
-        return getattr(obj, 'total_days', obj.days.count())
+        """Use annotated value if available, excluding Day 0 from the count."""
+        total = getattr(obj, 'total_days', obj.days.count())
+        if obj.has_day_zero:
+            total -= 1
+        return total
     
     def get_current_day_number(self, obj):
         """Use annotated value if available"""
         try:
-            offset = obj.start_day_number - 1
+            offset = -1 if obj.has_day_zero else 0
 
             # Use the annotation if available
             if hasattr(obj, 'current_day_count'):
@@ -289,7 +292,7 @@ class FastSerializer(serializers.ModelSerializer, ThumbnailCacheMixin):
             else:
                 return None
         except Exception as e:
-            print(f"An error occurred: {e}")
+            logging.error(f"Error computing current_day_number for Fast {obj.id}: {e}")
             return None
 
     def get_thumbnail(self, obj):
--- a/hub/migrations/0048_add_start_day_number_to_fast.py
+++ b/hub/migrations/0048_add_start_day_number_to_fast.py
@@ -12,8 +12,8 @@ class Migration(migrations.Migration):
     operations = [
         migrations.AddField(
             model_name='fast',
-            name='start_day_number',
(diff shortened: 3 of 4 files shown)


aylitalo-kog (commented):
Overall the approach is sound — adding a configurable start day to support "Day 0" fasts is the right idea. A few changes needed before merging.

hub/models.py
```diff
@@ -69,6 +69,11 @@ class Fast(models.Model):
         blank=True,
         help_text="Attribution or author of the culmination feast message"
     )
+    start_day_number = models.PositiveIntegerField(
+        default=1,
+        help_text="The day number assigned to the first day of the fast (e.g., 0 if the fast has a 'Day 0')",
+    )
```
aylitalo-kog: Switch this to a `BooleanField` instead of `PositiveIntegerField`. The only real use case is binary — either the fast starts at Day 0 or Day 1. An integer field invites nonsensical values like `start_day_number=5`.

Suggestion:
```python
has_day_zero = models.BooleanField(
    default=False,
    help_text="If True, the fast starts counting from Day 0 instead of Day 1",
)
```

Then in the serializer the offset becomes:
```python
offset = -1 if obj.has_day_zero else 0
```

hub/serializers.py
```diff
@@ -276,14 +276,16 @@ def get_total_number_of_days(self, obj):
     def get_current_day_number(self, obj):
         """Use annotated value if available"""
         try:
+            offset = obj.start_day_number - 1
+
             # Use the annotation if available
```
aylitalo-kog: `total_number_of_days` should exclude the 0th day when a fast has one. Right now `get_total_number_of_days` returns the raw `Day` row count. If a fast has a Day 0 + 39 real fasting days = 40 `Day` rows, the frontend sees `total_number_of_days=40` but `current_day_number` goes from 0 to 39 — the counter never reaches the total.

When `has_day_zero` is True, subtract 1 from the total so the progress display stays consistent (e.g., "Day 5 of 39" not "Day 5 of 40").

hub/serializers.py
```diff
@@ -276,14 +276,16 @@ def get_total_number_of_days(self, obj):
...
+                return obj.current_day_count + offset
+
             # Otherwise compute it
             days = obj.days.filter(date__lte=self.current_date).order_by('date')
             if days.exists():
-                return days.count()
+                return days.count() + offset
             else:
                 return None
         except Exception as e:
```
aylitalo-kog: This `print()` should be `logging.error()` (or `logging.exception()`) to match the rest of the codebase. `print()` won't appear in structured logs.

hub/serializers.py
```diff
@@ -276,14 +276,16 @@ def get_total_number_of_days(self, obj):
     def get_current_day_number(self, obj):
         """Use annotated value if available"""
         try:
+            offset = obj.start_day_number - 1
```
aylitalo-kog: This change needs tests. At minimum:

1. A test with `has_day_zero=False` confirming `current_day_number` is still 1-based.
2. A test with `has_day_zero=True` confirming the first day returns 0 and `total_number_of_days` excludes the zeroth day.
3. An edge case where no days have elapsed yet — should still return `None`.

PR #315, opened 2026-02-25
Fix duplicate Day records and add unique constraint on (date, church)

mattash:
## Summary

- **Root cause**: `create_combined_devotional` in `hub/admin.py` looked up `Day` objects by `(date, fast)` instead of `(date, church)`. Since the readings API creates Days by `(date, church)` with `fast=None`, the admin never found existing Days and silently created duplicates. Over time this caused `MultipleObjectsReturned` errors whenever `get_or_create(date, church)` encountered these rows.
- Fixed three `get_or_create` call sites in `admin.py` that were missing `church` from the lookup key.
- Added `UniqueConstraint(date, church)` to the `Day` model so the database enforces this invariant going forward.
- Migration `0049` includes a data migration step that cleans up the two existing duplicate rows in production before adding the constraint.

## Sentry issues resolved

| Issue | Description |
|-------|-------------|
| `PYTHON-DJANGO-K9` | `Day.MultipleObjectsReturned` on `/api/readings/` — direct symptom |
| `PYTHON-DJANGO-KJ` | `ProgrammingError` on `hub.tasks.update_current_fast_maps` |
| `PYTHON-DJANGO-KH` | Error updating current fast maps |
| `PYTHON-DJANGO-KR` | Error checking devotional availability |
| `PYTHON-DJANGO-KP` | Error checking fast beginning events |
| `PYTHON-DJANGO-KN` | `ProgrammingError` on `send_ongoing_fast_push_notification` |
| `PYTHON-DJANGO-KM` | Error checking completed fast milestones |
| `PYTHON-DJANGO-KK` | `ProgrammingError` on `send_fast_reminder_task` |
| `PYTHON-DJANGO-KS` | Error checking participation milestones |

Note: the KJ/KH/KR/KP/KN/KM/KK/KS issues were caused by a missing staging migration (`has_day_zero`) which has already been applied. Those Sentry issues have been resolved. K9 is fixed by this PR.

## Test plan

- [x] 882 tests passing (`python manage.py test --exclude-tag=performance --settings=tests.test_settings`)
- [ ] Run `manage.py migrate` on staging and production — the data migration will automatically clean up duplicate Day rows before adding the constraint
- [x] Verify `/api/readings/` no longer throws `MultipleObjectsReturned` — `readings.py` uses `get_or_create(date, church)`, now backed by the DB-level `unique_day_per_church` constraint that makes duplicate Days unrepresentable
- [x] Verify admin `create_combined_devotional` reuses existing Day records instead of creating duplicates — routed through the shared `get_or_create_days_for_fast()` helper keyed by `(date, church)`; covered by `test_combined_devotional_day_resolution.py` + `test_fast_admin_day_resolution.py`

---

> [!NOTE]
> **High Risk**
> Includes a data migration that deletes and rewires records to satisfy a new uniqueness constraint, so production data integrity depends on the dedupe logic behaving correctly and covering all references.
> 
> **Overview**
> Fixes duplicate `Day` creation by changing admin `Day.objects.get_or_create` lookups to key by `(date, church)` (and only set `fast` via `defaults`), including fast day-add/duplication flows.
> 
> Adds a DB-level `UniqueConstraint(date, church)` to `Day` and a migration that *deduplicates existing rows* by selecting a keeper per `(date, church)`, re-pointing `Reading`/`Devotional`/`Feast` FKs, then deleting extras.
> 
> Updates several tests to avoid creating overlapping `Day` rows for the same `(date, church)` now that the uniqueness invariant is enforced.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 96d481167a79b0f040407bf14f257fe70218c4b6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/events/tests/tests.py
+++ b/events/tests/tests.py
@@ -1999,12 +1999,13 @@ class UserMilestoneTasksTest(TestCase):
             year=2024
         )
         
-        # Create days for the weekly fast (ended yesterday)
+        # Create days for the weekly fast (ended 2 days ago; using a date different
+        # from setUp's yesterday to avoid the unique constraint on (date, church)).
         from django.utils import timezone
         from datetime import timedelta
-        
-        yesterday = timezone.now().date() - timedelta(days=1)
-        Day.objects.create(fast=weekly_fast, date=yesterday)
+
+        two_days_ago = timezone.now().date() - timedelta(days=2)
+        Day.objects.create(fast=weekly_fast, date=two_days_ago)
         
         # User participated in the weekly fast
         self.profile.fasts.add(weekly_fast)
--- a/hub/admin.py
+++ b/hub/admin.py
@@ -76,6 +76,31 @@ def _get_fk_links_url(fk_queryset, fk_name, max_num_to_show=_MAX_NUM_TO_SHOW):
     return format_html(format_string, *args)
 
 
+def get_or_create_days_for_fast(fast, dates):
+    """Get or create Days for ``fast`` keyed by (date, church).
+
+    Enforces the one-fast-per-(date, church) invariant backed by the
+    ``unique_day_per_church`` constraint. Reuses an existing Day for the date
+    (e.g. one auto-created by the readings API with ``fast=None``) so callers
+    can adopt it onto ``fast`` via ``fast.days.add()``/``.set()``. Returns
+    ``(days, conflicts)`` where ``conflicts`` are existing Days already owned
+    by a *different* fast; callers should surface these rather than silently
+    stealing the day from its current fast.
+
+    Shared by every admin flow that attaches Days to a Fast (add-days,
+    create-with-dates, duplicate, and combined-devotional creation) so they
+    all honor the invariant identically.
+    """
+    days = []
+    conflicts = []
+    for date in dates:
+        day, created = Day.objects.get_or_create(date=date, church=fast.church)
+        if not created and day.fast_id and day.fast_id != fast.pk:
+            conflicts.append(day)
+        days.append(day)
+    return days, conflicts
+
+
 @admin.register(Church, site=admin.site)
 class ChurchAdmin(admin.ModelAdmin):
     list_display = (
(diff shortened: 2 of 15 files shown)


andylitalo:
## Design decision worth stating explicitly: this gates one fast per church per day

`Day.fast` is a single FK, so a `Day` already belongs to at most one `Fast`. `unique(date, church)` promotes that to a hard DB invariant: **a church can have at most one fast covering any calendar date.**

That matches our domain — we don't run overlapping fasts within a single church — so I'm 👍 on keeping it. But it's a real semantic narrowing that's now baked into the schema, so let's document the decision:

- `FastByDateView` is a `ListAPIView` and `Fast.objects.filter(church=church, days__date=target_date)` is written to return *multiple* fasts. In practice it now returns ≤1 per church; the generic list shape hides that. If we ever want concurrent fasts in one church (e.g. weekly + seasonal), both this constraint and the `Day.fast` FK would need to change together.
- Several tests previously created overlapping fasts for the same church and had to be reworked to non-overlapping windows. That's the constraint doing its job, not a test smell — but it shows the behavior change is broad.

Recommend a one-line comment on the constraint in `hub/models.py` capturing this rationale so the next person doesn't have to reverse-engineer it.

andylitalo:
## Correctness: a fourth `Day` creation site is missed and will 500 once the constraint is live

`create_fast_with_dates` (`hub/admin.py`) still does a raw insert:

```python
Day.objects.create(date=date, fast=fast, church=data["church"])
```

The readings API auto-creates a `fast=None` `Day` for every browsed `(date, church)` (`hub/views/readings.py`), so after the constraint lands, creating a fast over any previously-browsed date raises `IntegrityError` → unhandled 500 in the admin. This PR fixed three `get_or_create` sites but left the one doing a bare insert — the most likely to collide.

Two related things in the same area:

1. `add_days_to_fast` / `duplicate_fast_with_new_dates` use `fast.days.add(*days)` / `.set(days)`. Because `Day.fast` is a reverse FK, those **re-point** a day's FK — i.e. silently steal a day (including another fast's day) rather than erroring. With the tighter `(date, church)` matching this is now more likely to fire.

**Recommended fix (I have a patch prepared):** route all three flows through one helper that `get_or_create`s by `(date, church)`, reuses/adopts an existing `fast=None` day, and returns any day already owned by a *different* fast as a conflict — which callers surface as an admin error (and roll back) instead of stealing it. `create_fast_with_dates` wraps the save+day resolution in `transaction.atomic()` so a conflict rolls the new fast back cleanly. Comes with admin tests for reuse / adoption / cross-church-non-conflict / conflict.

andylitalo:
## Migration 0049 is destructive by design — add observability before the prod run

`remove_duplicate_days` keeps one row per group and discards the rest's associations:

- an extra's `fast` is dropped when the keeper already has one;
- `Feast.objects.filter(day=extra).delete()` **destroys** the extra's feast (even a *different* one) when the keeper has any feast.

The reverse op is `RunPython.noop`, so this is irreversible. For our data it should only be merging accidental readings/devotional dupes (readings row `fast=None` + devotional row `fast=X` for the same logical day — exactly the root cause described above), but the migration currently runs blind.

**Recommended (patch prepared):**
1. Run a read-only audit query against prod first to confirm the duplicate groups are what we expect (accidental dupes, not distinct fasts/feasts).
2. Log every duplicate group and every lossy drop (dropped `fast` association, deleted feast) at WARNING so the effect is auditable in the migrate output / Sentry. Exact-duplicate reading/devotional deletes are safe (keeper already has an identical row) and don't need per-row logging.

Cheap insurance on a High-Risk, irreversible migration.

andylitalo:
## Scope + follow-ups

- **Unrelated icons migration.** `icons/migrations/0002_alter_icon_image.py` is unrelated to the Day fix (it captures a cosmetic `Icon.image` `help_text` change) and muddies review/rollback of a High-Risk data migration. Recommend pulling it into its own one-line PR. ⚠️ Note: removing it leaves a `makemigrations --check` drift on the `icons` app — if CI runs that check it will fail until the tiny follow-up PR lands.

- **Positive constraint test missing.** Nothing currently asserts the constraint actually fires — the touched tests only *avoid* tripping it. Worth one test that a second `Day.objects.create()` for the same `(date, church)` raises `IntegrityError`, so the invariant is guarded against regression. (Prepared.)

- **Pre-existing test failure on this branch (not caused by this PR).** `hub/tests/test_feast_designation.py` fails 1–4 tests depending on ordering — `feast.designation` state leaks between the LLM-designation test methods. Reproduces with these commits reverted, so it predates this PR, but the branch is currently red and the PR body's "882 tests passing" no longer matches the current head (893 tests, 1 failure in a full run). Worth fixing or tracking separately.

andylitalo:
### Correction on the icons migration point (from my follow-ups comment above)

I dug into this after removing `icons/migrations/0002_alter_icon_image.py`, and my earlier note about it leaving a `makemigrations --check` drift is **wrong** — I'd tested against this branch's stale base.

`main` already contains `icons/migrations/0002_alter_icon_image.py` (generated 2026-02-28), and it's **functionally identical** to the one this PR added (generated 2026-03-23) — same dependency, same `AlterField` on `icon.image`, same help_text; they differ only in quote style and the timestamp comment. In other words this branch independently generated a **duplicate migration with the same 0002 number** as one that landed on `main` after this PR branched — a numbering collision that's part of the staleness here.

So removing it is the **correct** resolution, and **no follow-up PR is needed**: once this branch is rebased/merged onto current `main`, main's `0002` provides the change, the model matches, and there's no drift. Please just make sure the rebase keeps `main`'s `0002_alter_icon_image.py` and drops this branch's copy (done in the pushed commit).

andylitalo (changes requested):
Requested some changes (noted in comments)

andylitalo (approved):
Both review comments are now addressed; approving.

**cursor #1 — migration FK-collision safety** (`0056_day_unique_date_church.py`): the dedup now skips/deletes rows that would violate `unique_reading_per_day` and the Devotional `(day, order, language_code)` constraint before re-pointing FKs, reassigns all Feasts (correct, since `unique_feast_per_day` was removed in `0052`), adopts the keeper's fast, and logs every merge/drop at WARNING. Resolved.

**cursor #2 — `create_combined_devotional` losing the fast association**: fixed in 8a51f07. All admin flows that attach Days to a Fast now route through the shared `get_or_create_days_for_fast` helper. `create_combined_devotional` adopts a reused `fast=None` Day onto the fast (so `day.fast` is set — restoring admin `day__fast` filtering and `DevotionalSet` cache invalidation) and, unlike the naive `fast.days.add(day)`, refuses to steal a Day already owned by a *different* fast, surfacing an admin error and rolling back. Covered by `hub/tests/test_combined_devotional_day_resolution.py` (adoption, new-Day, and conflict cases; the adoption test was verified to fail without the fix). Resolved.

CI green (`test (3.11)`), 484 hub/unit tests pass locally.

andylitalo:
@mattash do we still want to merge these? Just want to avoid them getting stale.

PR #460, opened 2026-07-16
Offline Eastern Armenian Bible corpus (BibleVerse) to retire sacredtradition.am scraping

andylitalo:
## Goal

Give bahk a **local, offline, verse-addressable Eastern Armenian Bible** so we can stop scraping `sacredtradition.am` at runtime for Armenian (`hy`) reading text. Today `hub/services/reading_text_service.py:fetch_armenian_text` hits the site live for every reading; readings vary year to year, so the fix is to store the whole canon once and compose passages from `armenian_lectionary` output.

**Scope: this PR provides the tool only.** It does **not** rewire `fetch_armenian_text` — cutting the dependency over is a separate, deliberate follow-up once these building blocks are in place. Nothing in the serving path changes yet, so this is safe to land and deploy incrementally.

## Source

The **1994 Nor Ejmiatsin (Նոր Էջմիածին) Eastern Armenian translation**, as published in the `sacredtradition.am` online library (`Library/getBible.php?idArt=…`, books 2801–2875, 75 books — full canon incl. deuterocanon). A **one-time scrape** captured all 75 books into git-tracked, verse-keyed per-book JSON; after that snapshot **no further network calls are made** — every correction below is re-derived from the cached raw HTML.

## What's here

- `scripts/scrape_bible_hy.py` — one-time scraper → verse-keyed per-book JSON + validation report. Supports fully offline re-parse (`--from-dir`) so fixes never require the network.
- `hub/data/bible_hy/` — `usfm_mapping.json` (75-row idArt→USFM), `books/<idArt>.json` (source of truth), plus `errata.json` / `ERRATA.md` documenting every correction.
- `hub/models.py:BibleVerse` (+ migration, admin) — `(version, book=USFM, chapter, verse, text)`, superscriptions as verse 0. Retrieval via `passage_queryset` / `compose_passage` (inline `[v]` markers matching the current format).
- `hub/management/commands/load_bible_hy.py` — full-rebuild loader into Postgres/sqlite.

## Errors captured in the errata, fixed in the scrape

The source has real defects (mislabeled verses, dropped chapter markers, LXX merges/reorders). Rather than hand-edit the JSON, each fix is **encoded in `errata.json` and applied deterministically during the scrape**, and any *unresolved* duplicate verse number is a **hard error** so nothing is ever silently overwritten. Lectionary-served fixes were validated byte-for-byte against the daily-readings page:

[...]

PR #461, opened 2026-07-17
Retire sacredtradition.am for readings: lectionary engine + offline corpus

andylitalo:
## Goal

Stop scraping `sacredtradition.am` at runtime for **daily readings**. Reading *references* now come from the offline `armenian_lectionary` engine, and Armenian *verse text* from the offline `BibleVerse` corpus added in #460. After a one-time DB load, no runtime sacredtradition call remains for readings.

**Stacked on #460** (`feat/armenian-bible-corpus`) — it depends on the `BibleVerse` model, migration `0054`, and the corpus data introduced there. Merge #460 first (or merge this into that branch).

## What changed

| Concern | Before | After |
|---|---|---|
| Reading references | `hub/utils.py::scrape_readings` (scrape) | `hub/services/lectionary_service.py::get_daily_readings` → `armenian_lectionary` (offline) |
| Armenian verse text | `scrape_armenian_reading_texts` (scrape) | `fetch_armenian_text` → `BibleVerse.compose_passage` (offline corpus) |
| `book_hy` | byproduct of the Armenian scrape | recovered from version-controlled `usfm_mapping.json` |
| English text | API.Bible | unchanged |
| Feast names | `scrape_feast` | **unchanged** (sole remaining sacredtradition call) |

- `get_daily_readings` reuses the existing `PARSER_REGEX` to parse the engine's citation strings into the exact dict shape `scrape_readings` returned (drop-in). Book names already exist in `BOOK_NAME_TO_USFM` / `CATENA_ABBREV_FOR_BOOK`, so USFM lookup and catenabible URLs work unchanged.
- Trust the engine fully within the validated **2001–2027** window (configurable via settings); unsupported churches / out-of-range dates → `[]`.
- Deleted the two dead scrapers (−245 lines in `utils.py`); dropped the Armenian batch pre-scrape resource preparer.

## Decisions (per discussion)
- **Feasts untouched.** The engine's `"Liturgical Day"` is English-only with data defects (adjacent commemorations concatenated; a stray Cyrillic char) and has no `name_hy`, so `scrape_feast` stays. Fully eliminating sacredtradition is a follow-up once the engine gains clean feast names.
- **Confidence gating:** trust all readings in-range; no readings out-of-range (engine returns empty for underivable days anyway).

## Testing
- Rewrote the Armenian-text tests to exercise the `BibleVerse` corpus; added `test_lectionary_service.py` (parsing, year-guard, unsupported church, real Easter-2026 readings); repointed mock targets in affected tests. All touched modules pass; `ruff` clean.
- **Real end-to-end (offline):** loaded the full 35,739-verse corpus, computed Easter 2026's 14 readings, and composed the actual Armenian text for John 20:1-18 (`[1] Եւ կիրակի օրը Մարիամ Մագդաղենացին…`, `book_hy = "Աւետարան ըստ Յովհաննէսի"`). No network calls.

## ⚠️ Before merge / deploy
- **Licensing:** this makes the 1994 Nor Ejmiatsin text user-facing — resolve the permission/attribution question flagged in #460 first.
- **One-time data load:** apply migration `0054_bibleverse` and run `python manage.py load_bible_hy` in each environment before the Armenian-text cutover has data to read.


mattash:
armenian_lectionary 1.1.0 returns `Malach` for Malachi, but Bahk’s downstream mappings recognize only `Malachi`. Persisting the value unchanged causes fresh Malachi readings to fail English/offline Armenian text lookup, and Catena falls back to its home page.

Please normalize `Malach` to `Malachi` before persistence (or add `Malach` aliases to all relevant mappings), and add an integration test using a date that returns `Malach`.

andylitalo:
Addressed in 0ae94d3. Rather than normalizing `Malach` downstream, this is now fixed at the source: **armenian-lectionary 1.2.2** (published to PyPI) corrects the typo upstream — the Presentation-eve (Feb 13) block ships the canonical `Malachi 3.1-4` everywhere, and the stale `Malach` key was dropped from the `hy` book-name map. So the engine returns `Malachi` directly and Bahk's existing mappings resolve it with no normalization layer needed.

- Bumped the pin: `armenian-lectionary>=1.2.2`.
- Added an integration test (`test_presentation_eve_malachi_spelling`) on Feb 13, 2026 asserting the book is `Malachi` (and never `Malach`), with the full 3:1–4 reference.

All 12 `test_lectionary_service` tests pass locally against 1.2.2.

andylitalo:
Also @mattash, how should we handle licensing the Nor Ejmiatsin translation of the Bible? This PR will have us serving it to the frontend. Is there a Copyright we should add? Or stick to what we did when we scraped sacredtradition.am?

mattash:
> Also @mattash, how should we handle licensing the Nor Ejmiatsin translation of the Bible? This PR will have us serving it to the frontend. Is there a Copyright we should add? Or stick to what we did when we scraped sacredtradition.am?

I'm going to contact Surpazan about this for guidance.

mattash (commented):
There is still a production-path mismatch around Armenian book names. `get_daily_readings()` returns only `book`/`book_en` plus the range fields, but `import_readings` only persists an Armenian book name when an optional `book_hy` key is present. The updated command test continues to mock `book_hy`, even though the real lectionary service never produces it, so the test masks the regression.

As a result, readings created by `import_readings` retain an empty `book_hy` and display the English fallback in Armenian until a separate text-fetch/backfill happens. Please either populate `book_hy` from `usfm_mapping.json` in the production import path or explicitly run the same mapping/backfill during persistence, and update the test to use the real output shape from `get_daily_readings()`.

andylitalo:
Good catch — fixed in 3bf53e0.

Root cause was exactly as described: `get_daily_readings()` never returns a `book_hy` key (only `book`/`book_en` + range fields), so `reading.pop("book_hy", None)` always resolved to `None` in production, and the mocked tests (which included a `book_hy` key that real output never has) hid it.

Rather than run a separate mapping/backfill step, I resolved `book_hy` from `usfm_mapping.json` directly at persistence time — added `reading_text_service.book_hy_for_book(book_en)`, which reuses the same USFM lookup `fetch_armenian_text()` already relies on for its own backfill, and called it in `import_readings.py` where you flagged it. `GetDailyReadingsForDate` in `views/readings.py` had the identical dead-key pattern (it happened to self-heal a moment later via the synchronous text fetch, but was still briefly wrong), so I fixed that too for consistency.

Updated `test_import_readings.py` and the `views/readings.py` translation test to mock the real `get_daily_readings()` output shape (no `book_hy` key) and assert against the actual `usfm_mapping.json` values. Also swapped the "no translation" case from `Matthew` (which *is* in the mapping and now resolves automatically — `Աւետարան ըստ Մատթէոսի`) to `Azariah`, which genuinely has no `BOOK_NAME_TO_USFM` entry, so that test still exercises the `None` path honestly.

All 439 `hub` tests pass locally; `ruff` clean.

I also checked the other open PRs against this branch: #467 (stacked on this one) relocates the same buggy line into a new `persist_readings()` helper without fixing it, and #468 doesn't touch `import_readings.py` at all — so neither would have masked or duplicated this fix, and #467 will just need a rebase to pick it up.

mattash (approved):
Surpazan advises that we should proceed with using the text, and that copyright would only come into effect in the case that we were print publishing the bible text. Therefore, we should move forward with this and related revisions.

andylitalo:
> Surpazan advises that we should proceed with using the text, and that copyright would only come into effect in the case that we were print publishing the bible text. Therefore, we should move forward with this and related revisions.

Thank you for checking Der Hayr! I'll merge. We will just cite the source.

PR #465, opened 2026-07-25
Persist and source the Azariah composite reading (Song of the Three)

andylitalo:
## Goal

Stop dropping the **Azariah** half of the engine's one composite reading. The `armenian_lectionary` engine emits `"Daniel 3.1-23, Azariah. 1-68"` (the *only* scripture composite it produces, ~9 days/range); the reader previously did `split(",")[0]` and silently discarded the Prayer of Azariah / Song of the Three. This persists **both** sub-references as separate `Reading`s and gives Azariah real English and Armenian text.

> **Stacked on #461** (`feat/retire-sacredtradition-readings`) — it modifies `lectionary_service.py`, which #461 introduces. Merge #461 first (or merge this into that branch).

## What changed

| Concern | Before | After |
|---|---|---|
| Composite reading | first sub-ref only (Azariah dropped) | one `Reading` per sub-reference (Daniel **and** Azariah) |
| Azariah → English | n/a (dropped) | `"Azariah" → S3Y`, served from API.Bible **KJVAIC** like other apocrypha |
| Azariah → Armenian | n/a (dropped) | standalone `S3Y` book (`name_hy` **Աղօթք Ազարիայ**) derived from Daniel 3:24-90 |

- **Split** (`hub/services/lectionary_service.py`): comma-composites now split into one reading per sub-reference; the stray trailing period the engine emits on `"Azariah."` is stripped so the book resolves (`normalize_book_name` does not strip periods).
- **English** (`hub/constants.py`): `"Azariah" → "S3Y"` (+ aliases). `S3Y` is already in `APOCRYPHA_USFM_IDS`, so text routes to KJVAIC.
- **Armenian** (corpus): the Nor Ejmiatsin corpus has **no standalone `S3Y`** — the Prayer of Azariah + Song of the Three is embedded in **Daniel 3:24-90** (Armenian Daniel 3 runs to 97 verses). `scripts/derive_azariah_hy.py` extracts it into a standalone book renumbered from 1 (`hub/data/bible_hy/books/2876.json`, synthetic idArt after the scraped 2801-2875), with a `usfm_mapping.json` row giving `name_hy = Աղօթք Ազարիայ`. English and Armenian now compose the same `S3Y` id. `load_bible_hy` `EXPECTED_BOOKS` 75 → 76 (dry-run: S3Y = 67 rows).

The heads align exactly against the corpus text: HY `Daniel 3:24` = EN `S3Y` v1 ("they walked in the fire, praising"), `3:25` = v2 ("Azariah stood and prayed").

## Testing
- `hub.tests.test_lectionary_service`: composite splits into Daniel 3:1-23 **and** Azariah 1-68; trailing period stripped; Great Saturday 2026 (real offline engine) persists both.
- `hub.tests.test_armenian_text`: Azariah composes from the standalone `S3Y` corpus book with `book_hy = Աղօթք Ազարիայ`.
- All touched modules pass (77 tests across the three suites); `ruff` clean.

## ⚠️ Before merge / deploy
- **One-time data load:** run `python manage.py load_bible_hy` in each environment to ingest the new S3Y rows.
- **Known follow-up (2026-08-04, when API.Bible access returns):** the derived book is flagged `needs_review` because Armenian is **67** verses vs English `S3Y`'s **68** — KJV splits one doxology verse that Armenian keeps whole. Fetch the English S3Y text to pin the exact split verse, then finalize the tail numbering in `derive_azariah_hy.py`. Also confirm KJVAIC returns `S3Y` 1-68 (couldn't verify now — the dev key 403s even on Tobit).


scripts/derive_azariah_hy.py
```diff
@@ -0,0 +1,80 @@
...
+book renumbered from 1, letting English (KJVAIC ``S3Y``) and Armenian compose the same book id.
+
+Boundaries (verified against the corpus text):
+  * Daniel 3:23 -- the three fall bound into the furnace (end of the Hebrew narrative).
+  * Daniel 3:24 -- "they walked in the fire, praising"  == KJV S3Y verse 1.
+  * Daniel 3:25 -- "Azariah stood and prayed"           == KJV S3Y verse 2.
+  * Daniel 3:90 -- "all ye servants of the Lord, bless" == end of the Song.
+  * Daniel 3:91 -- Nebuchadnezzar's reaction (Hebrew narrative resumes).
+
+So HY verse N = Daniel 3 verse (N + 23), giving S3Y 1-67.  NOTE: English ``S3Y`` is **68** verses;
```
mattash, line 16: **[P1] Reconcile the 67-versus-68 mapping before shipping this corpus.** The persisted reading is Azariah 1–68, but this derivation writes only 67 rows and marks the book `needs_review`. `compose_passage()` treats any non-empty subset as success, and the new test only seeds verses 1, 2, and 67, so the API can report `endVerse: 68` while the Armenian text ends at `[67]`. This is not merely one tail split: English S3Y 66 maps across the derived Armenian rows 63–65, followed by English 67→HY 66 and 68→HY 67, leaving the verse labels misaligned. Please establish the full source mapping, derive the correct labels/content, clear `needs_review`, and add a complete 1–68 regression test before loading it.

mattash (changes requested):
Requesting changes on the Armenian S3Y versification. The PR currently ships a 67-row derived corpus for a persisted 1–68 reading, while the added test accepts sparse/partial text as success. Please reconcile the mapping and validate the complete passage before loading this corpus.

andylitalo:
Good catch. The issue was the complexity of the verse-mapping. This is a general issue across the Bible (Esther, Romans ch. 14, Daniel) where verses don't align between KJV and Armenian versification. Closing this patch for a more general solution (upcoming).

PR #466, opened 2026-07-25
Bound API.Bible spend and stop the weekly quota burn

andylitalo:
## Problem

The weekly refresh task was making **~1700 API.Bible calls every Monday** against a 5,000/month quota.

**Cause: a retry spiral.** A failed fetch never records `text_fetched_at` — `fetch_english_text` catches the exception and returns `False`. So once the quota was gone and API.Bible began returning 429/403, every reading stayed permanently stale, and each Monday the task retried all of them and failed all of them. Nothing ever recovered on its own.

Three pieces of production data pin this down:

- **Spend is spiky on Mondays with nothing the rest of the week.** The burn is the Beat job, not continuous view traffic.
- **`hub_reading` held exactly 2000 rows on a Saturday.** The prune runs Monday and trims *to* `MAX_READINGS`=2000, so a mid-week count of exactly 2000 proves **zero rows were created all week** — no crawler-driven creation, and no on-demand fetching either.
- **Every Monday, not every fourth.** Successful fetches would stamp all 2000 with that Monday's timestamp, and the 23-day threshold would make the next three runs no-ops. Weekly recurrence is only possible if the timestamp is never written.

Worth stating plainly for the record: **this was not crawler-driven.** An earlier reading of the code blamed the prune/re-create churn plus crawler traffic on the public `/readings/<date>` page. `robots.txt` shipped in bahk_landing#41 on 2026-07-04 and changed spend by nothing over the following three weeks, and the row-count evidence above rules the churn story out. Crawlers may have been the original ignition months ago; they are not what has been burning the quota. (bahk_landing#45, which added CDN caching for the same theory, was closed.)

The prune is still removed here — it deletes *today's* readings first while keeping far-future ones, which is backwards — but it was not the cause.

## Changes

- **Stop deleting readings.** The prune deleted *today's* readings before far-future ones and forced rows to be re-created and re-fetched later. The table grows instead.
- **Refresh the nearest `READING_REFRESH_LIMIT` (2000) stale readings per run**, today/upcoming ascending then past descending. Nobody browses far into the past, so this spends the allowance on dates people actually open.
- **Circuit breaker**: abort after `READING_REFRESH_MAX_CONSECUTIVE_FAILURES` (10) consecutive English failures that actually reached the API. A rejecting API now costs ~10 calls instead of ~1700. **This is the direct fix for the observed burn.** 10 rather than 1–3 so a transient blip or a single passage API.Bible 404s on cannot halt the whole refresh; 10 wasted calls/week is ~43/month, under 1% of quota.
  - Book names that fail `resolve_reading_passage` are excluded from the count. They return before any HTTP request and can never succeed, so they are a data problem, not evidence the API is unhealthy. If any exist they stay permanently stale and get re-selected every run, so counting them would abort the task once everything else is fresh. They are reported separately at WARNING so the fix (extend `BOOK_NAME_TO_USFM`) is discoverable.
  - Audited: across 6,057 readings spanning four years of the lectionary cycle, **all 62 distinct book names resolve** and no map key is unreachable after normalisation. So this guard is expected to report zero on current data. `main` builds readings from the sacredtradition.am scraper rather than the lectionary engine, though, so the guard is kept for names the scraper may produce that the engine does not.
- **`BIBLE_API_MONTHLY_BUDGET` (4500)**: a hard ceiling across *both* the refresh task and the on-demand view path, so no failure mode can exhaust the quota. Plus `READING_FETCH_DAILY_BUDGET` (75) for the public path alone. Cache-backed (`hub/services/api_budget.py`), fails closed.
- **On-demand re-fetch**: the view now refreshes text past `READING_TEXT_MAX_AGE_DAYS`, not just newly created readings — bounded by those budgets.
- **Serve-time freshness guard**: `get_reading_text_fields` blanks English text once it exceeds 30 days, so nothing past API.Bible's cap is served when a refresh doesn't reach a reading. Armenian is not API.Bible-sourced and never expires.

Spend budget: ~2,170/month from the weekly refresh + ~2,250 from the daily cap, against a 4,500 ceiling and the 5,000 quota.

Also fixes a telemetry bug (`api_calls` only counted readings whose *Armenian* composition also succeeded) and stops `resolve_reading_passage` failures from being logged as API errors.

## Behaviour worth knowing (not a bug)

Once the table exceeds ~2000 stale rows in a 23-day window, far-from-today readings will never be reached by the weekly task. Their text passes 30 days, gets blanked at serve time, and is re-fetched on demand the next time someone actually requests that date. That is the designed interaction between the refresh limit and the serve-time guard.

## Deploy

- **Unset `MAX_READINGS`** in production config vars (removed; set `READING_REFRESH_LIMIT` only if it had been overridden).
- Invariant: `READING_TEXT_REFRESH_DAYS` (23) must stay `<=` `READING_TEXT_MAX_AGE_DAYS` (30).
- Watch Sentry for `"API.Bible ... budget ... exhausted"` and `"Aborting refresh after N consecutive"` — both are env-tunable without a deploy.

## Testing

`hub.tests.test_bible_api` + `tests.unit.hub.test_fums`: **83 passing**, including new coverage for nearest-first selection, past-date backfill, never-deleting, the circuit breaker, monthly-ceiling cutoff, on-demand refetch at 31 vs 29 days, budget exhaustion leaving the row created with blank text, unmappable books not burning budget, and Armenian never expiring.

Full suite: 1310 tests, 1 error — `learning_resources.cleanup_orphaned_bookmarks_async` needs the Celery result backend at hostname `redis:6379`, which only resolves inside the devcontainer. Unrelated to this change; worth a re-run in CI to confirm.

One existing fixture was corrected: `test_readings_api_includes_fums_token` set `text`/`fums_token` with no `text_fetched_at`, a state production cannot produce (they're written in the same `.update()`).

### DIFF
--- a/hub/services/reading_text_service.py
+++ b/hub/services/reading_text_service.py
@@ -100,6 +100,7 @@ def fetch_english_text(
     *,
     service: BibleAPIService | None = None,
     budgets: list[APIBudget] | None = None,
+    stats: dict[str, int] | None = None,
     **_kwargs,
 ) -> bool:
     """Fetch English Bible text from API.Bible for a single Reading.
@@ -115,6 +116,11 @@ def fetch_english_text(
                  consumed only once the passage has resolved and immediately before
                  the call, so an unmappable book name never burns a token.  If any
                  budget refuses, no call is made and this returns False.
+        stats: Optional shared counter dict.  ``stats["attempted"]`` is incremented
+               once per call that actually reaches API.Bible — i.e. after an unmappable
+               book or an exhausted budget has already returned, so callers can tell
+               "we tried and it failed" apart from "we never tried" using only the
+               boolean return value.
 
     Returns:
         True if text was successfully fetched, False otherwise.
@@ -152,6 +158,9 @@ def fetch_english_text(
                 budget.period, budget.limit, reading.pk, reading.passage_reference,
             )
             return False
+
+    if stats is not None:
+        stats["attempted"] = stats.get("attempted", 0) + 1
 
     try:
         result = service.get_passage(
--- a/hub/services/reading_text_service.py
+++ b/hub/services/reading_text_service.py
@@ -74,6 +74,27 @@ def book_hy_for_book(book_en: str) -> str | None:
 #  Individual language fetchers
 # ------------------------------------------------------------------ #
 
+def reading_is_mappable(reading) -> bool:
+    """True when the reading's book resolves to a USFM id, i.e. a fetch would reach the API.
+
+    Lets callers tell "API.Bible refused us" apart from "we cannot name this book".  The
+    latter fails before any HTTP request, so it is a data problem to fix in
+    ``BOOK_NAME_TO_USFM`` — not a signal that the API is unhealthy.  Pure dict lookups,
+    no I/O, so it is cheap to call alongside a fetch.
+    """
+    try:
+        BibleAPIService.resolve_reading_passage(
+            reading.book,
+            reading.start_chapter,
+            reading.start_verse,
+            reading.end_chapter,
+            reading.end_verse,
+        )
+    except ValueError:
+        return False
+    return True
+
+
 def fetch_english_text(
     reading,
     *,
--- a/hub/services/reading_text_service.py
+++ b/hub/services/reading_text_service.py
@@ -8,22 +8,31 @@ them all for a given Reading.  Adding a new language is a three-step process:
     3. (Optional) Register a resource preparer in ``RESOURCE_PREPARERS`` if the
        fetcher benefits from batch-level shared state (e.g. a scraped page or
        HTTP session that can be reused across multiple readings).
+    4. (Optional) Register the language in ``EXPIRING_LANGUAGES`` if its source
+       licence caps how old served text may be.
 
 The view calls ``prepare_shared_resources`` once per batch, then
 ``fetch_all_reading_texts`` once per reading.  ``get_reading_text_fields``
 resolves model fields for the API response without hard-coding language names.
+
+``prepare_shared_resources`` also carries the API.Bible spend budgets, and only the
+readings view calls it \u2014 so the budgets gate the public on-demand path and nothing else.
+The weekly refresh task builds its own shared dict and charges only the monthly budget.
 """
 
 import json
 import logging
+from datetime import timedelta
 from functools import lru_cache
 from pathlib import Path
 from typing import Any
 
 from django.conf import settings
+from django.db.models import Q
 from django.utils import timezone
 
 from hub.constants import BOOK_NAME_TO_USFM_NORMALIZED, normalize_book_name
+from hub.services.api_budget import DAY, MONTH, APIBudget
 from hub.services.bible_api_service import BibleAPIService
 
 logger = logging.getLogger(__name__)
@@ -65,7 +74,13 @@ def book_hy_for_book(book_en: str) -> str | None:
 #  Individual language fetchers
 # ------------------------------------------------------------------ #
 
-def fetch_english_text(reading, *, service: BibleAPIService | None = None, **_kwargs) -> bool:
+def fetch_english_text(
+    reading,
+    *,
+    service: BibleAPIService | None = None,
+    budgets: list[APIBudget] | None = None,
+    **_kwargs,
+) -> bool:
     """Fetch English Bible text from API.Bible for a single Reading.
 
     Each Reading gets its own API call so that it receives a unique FUMS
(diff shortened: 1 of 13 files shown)


hub/services/reading_text_service.py
```diff
@@ -62,6 +105,22 @@ def fetch_english_text(reading, *, service: BibleAPIService | None = None, **_kw
...
             reading.end_verse,
         )
+    except ValueError as exc:
+        logger.error(
+            "Book name mapping failed for Reading %s ('%s'): %s",
+            reading.pk, reading.book, exc,
+        )
+        return False
+
+    for budget in budgets or ():
```
mattash, line 154: **[P1] Enforce the monthly ceiling across every fetch path.** Treating `budgets` as optional leaves existing callers unmetered: `fetch_reading_text_task` calls `fetch_english_text(reading)` without a budget, and the admin single/bulk actions still call the legacy `fetch_text_for_reading` helper directly. An ad-hoc backfill or bulk admin action can therefore make an arbitrary number of calls outside the counter and exhaust the same quota this PR is intended to protect. Please enforce the monthly budget at the lowest shared API-call layer, or update every entry point and cover the task/admin paths with regression tests.

hub/tasks/bible_api_tasks.py
```diff
@@ -57,81 +59,122 @@ def fetch_reading_text_task(self, reading_id: int):
...
+        # selects only such readings — the steady state once everything else is fresh —
+        # would trip the circuit breaker every week and cry wolf.
+        mappable = reading_is_mappable(reading)
+        if not mappable:
+            unmappable += 1
+
         results = fetch_all_reading_texts(reading, **shared)
-        if all(results.values()):
+
+        if results.get("en"):
```
mattash, line 134: **[P2] Count API attempts independently from successful fetches.** This increments `api_calls` only when the English fetch succeeds, even though the budget is consumed and `get_passage()` is called before a 429/403 failure is returned. A breaker trip after ten rejected requests will consequently log zero API calls. The attempted-count expression below can also double-count a reading whose English fetch succeeds but Armenian fetching fails, because it contributes to both `api_calls` and `failures`. Please track readings attempted, English API attempts, successes, and per-language failures separately so the incident telemetry reflects the calls that actually reached API.Bible.

mattash (changes requested):
Requesting changes on two gaps in the API.Bible spend controls: not every fetch path is covered by the monthly ceiling, and the new telemetry reports successful fetches rather than actual API attempts.

hub/services/reading_text_service.py
andylitalo: Fixed in 67db7da. Went with the "update every entry point" option rather than pushing the check into `get_passage()`: the existing tests mock `BibleAPIService.get_passage` directly to fake the HTTP call while exercising real budget logic (e.g. `test_refresh_stops_when_monthly_budget_is_exhausted`), so enforcing at that layer would've made those mocks silently bypass the check.

Instead, consolidated onto `fetch_english_text` as the one shared layer:
- `hub/admin.py`'s single/bulk fetch actions now call `fetch_english_text(reading, service=service, budgets=bible_api_budgets(include_daily=False))` instead of the separate `fetch_text_for_reading` helper.
- `fetch_reading_text_task` (the ad-hoc backfill task) now passes the same budgets — previously it called `fetch_english_text(reading)` with none.
- Deleted `fetch_text_for_reading` from `bible_api_service.py` entirely (and its now-unused `timezone` import) so there's no second, unbudgeted path left to forget.

Added `AdminFetchBudgetTests` in `hub/tests/test_bible_api.py` covering both admin entry points stopping once `BIBLE_API_MONTHLY_BUDGET` is exhausted.

hub/tasks/bible_api_tasks.py
andylitalo: Fixed in 67db7da. `refresh_all_reading_texts_task` now tracks four things separately instead of overloading `api_calls`:

- `readings_processed` — incremented once per loop iteration, replacing the old `api_calls + len(failures)` proxy that double-counted a reading which succeeded in English but failed in Armenian.
- `english_attempts` — a new `stats["attempted"]` counter threaded through `fetch_english_text` (optional `stats` kwarg), incremented right after the budget/unmappable checks pass and immediately before the actual `get_passage()` call — i.e. exactly the calls that reach API.Bible, whether they succeed or get rejected.
- `successes` — the old "increment on `results.get('en')`" logic, renamed now that it sits next to `english_attempts`.
- `failures_by_lang` — a per-language failure count alongside the existing per-reading `failures` list.

Both the abort-path log and the final summary log now report all four, so a breaker trip after N rejected calls shows N attempts and 0 successes instead of "0 API calls."

Added two regression tests: one asserting a breaker trip logs a non-zero attempt count, and one asserting a reading with an English success + Armenian failure is counted once in `readings_processed`, not twice.

mattash (approved):
Reviewed the current head. The monthly-budget coverage and API-attempt telemetry findings are addressed, the related conversations are resolved, and exact-head CI is green. Approving.

andylitalo:
> Reviewed the current head. The monthly-budget coverage and API-attempt telemetry findings are addressed, the related conversations are resolved, and exact-head CI is green. Approving.

Thanks @mattash! I'm going to merge: could you remove the now obsolete `MAX_READINGS` from production config vars? Also, could you add `READING_REFRESH_LIMIT=2000` for visibliity (2000 is the default value, but this way we can easiliy adjust the value if needed).

PR #467, opened 2026-07-25
Fix random reading order: assign sequence from lectionary engine

### LINKED ISSUE #324: readings sometimes show up out of order
image
image

andylitalo:
## Goal

Fix #324 ("readings sometimes show up out of order"). #461 replaced the sacredtradition.am scrape with the offline `armenian_lectionary` engine, but the engine's order was never persisted or enforced downstream, so the bug remains after #461.

**Stacked on #461** (`feat/retire-sacredtradition-readings`) — this PR is based on that branch so the diff shows only the ordering fix. Merge/rebase after #461.

## Root cause

`hub.services.lectionary_service.get_daily_readings` (added in #461) returns readings in the engine's order, but nothing preserved that order once persisted:

- `Reading` had no `sequence`/order field and no `ordering` in its `Meta` (unlike `Devotional`, which has both an `order` field and `ordering = ["day__date", "order"]`).
- `hub/views/readings.py` and `import_readings.py` both serialized/queried via `day.readings.all()` with no `order_by`.

So the served order was whatever order the rows happened to be stored in — fine only if rows are always created in engine order and never touched again. It breaks whenever rows already exist in a different order: a prior import, a retry, or a race between concurrent first-read requests (`day.readings.exists()` check + `get_or_create` isn't atomic against concurrent requests for the same day).

## Change

- **Migration `0056`**: adds `Reading.sequence` (nullable `PositiveIntegerField`) and sets `Reading.Meta.ordering = ["day__date", "sequence", "id"]`.
- **`hub/services/lectionary_service.py`**: new `persist_readings(day, readings)` helper — centralizes the `get_or_create` + translation-backfill logic previously duplicated in `import_readings.py` and `views/readings.py`, and assigns `sequence` from each reading's position in the (already-ordered) `readings` list. Assignment happens whether the row is newly created or already existed, so re-running the importer over a date range self-heals any previously out-of-order rows for that day.
- **`hub/management/commands/import_readings.py`** and **`hub/views/readings.py`**: both now call `persist_readings` instead of inlining the same logic.

## Deployment

Migration `0056` adds `Reading.sequence` but leaves it `NULL` on every existing row, and `persist_readings()` only assigns it when readings are freshly imported for a `Day` — a `Day` that already has readings never gets touched. So **already-out-of-order days stay broken after this deploys** unless the existing rows are backfilled once.

`hub/management/commands/backfill_reading_sequence.py` (added after review) does that backfill. **Run it once, after applying migration `0056`, before or alongside deploying this PR:**

```bash
python manage.py backfill_reading_sequence --strict
```

- It's idempotent and safe to re-run — matches existing `Reading` rows to the engine's current order by `(book, start_chapter, start_verse, end_chapter, end_verse)` and updates only `sequence`; it never creates or deletes rows.
- `--strict` exits nonzero if any `Day` has rows it couldn't match one-to-one (e.g. a book name the engine has since renamed) — worth doing on first run so any real mismatches surface instead of being silently skipped.
- Add `--dry-run` first if you want to preview the update counts before writing.
- Optionally scope with `--church "<name>"` / `--start-date YYYY-MM-DD` / `--end-date YYYY-MM-DD` to backfill incrementally instead of the whole table at once.

This is a manual step for now — there's no `release:` process in the `Procfile` to hook it into automatically.

## Test plan

Added `PersistReadingsOrderTests` in `hub/tests/test_lectionary_service.py`:
- `test_preexisting_rows_out_of_order_stay_out_of_order_without_sequence` — reproduces the bug directly: creates `Reading` rows for a day in the *reverse* of engine order (simulating pre-existing rows from an earlier import/race), then shows that querying by creation order (the only signal available pre-fix) does **not** match engine order.
- `test_persist_readings_assigns_sequence_matching_engine_order` — same out-of-order setup, but runs it through `persist_readings`; `day.readings.all()` (relying only on `Reading.Meta.ordering`, exactly as the view/command do) now matches engine order.
- `test_persist_readings_on_fresh_day_matches_engine_order` — sanity check for the common case (no pre-existing rows).

Added `hub/tests/test_backfill_reading_sequence.py` for the backfill command: assigns `sequence` to reversed pre-existing rows, `--dry-run` makes no changes, a second run is a no-op, unmatched rows are reported (and left untouched, not created/deleted), `--strict` raises on a mismatch, and `--church`/date-range scoping only touches the intended `Day`s.

Also ran the full `hub` suite (442 tests, `--exclude-tag=performance`) and `ruff check` on all touched files — all pass. `makemigrations --check` shows only pre-existing, unrelated drift (icons app).


mattash (commented):
The new `sequence` field and `persist_readings()` logic fix ordering for readings that pass through that helper, but deployment does not repair existing rows: the migration leaves every existing `sequence` as `NULL`, and the request view calls `persist_readings()` only when `day.readings.exists()` is false.

Please add a dedicated, idempotent management command to backfill existing reading sequences after the migration. A pure backfill should compute the lectionary order for each existing `Day`, match existing rows by book plus chapter/verse range, update only matched rows, and avoid silently creating or deleting readings. It should support date/church scoping, `--dry-run`, and a strict mode that exits nonzero while reporting unmatched database or engine readings. Processing each day transactionally would keep it restartable and limit rollback scope.

Please also add tests covering reversed existing rows, dry-run behavior, idempotency, and mismatch handling, and document the command as a mandatory deployment step after applying the migration. This keeps engine-dependent backfill logic out of the schema migration while ensuring the PR actually fixes already-existing out-of-order data.

andylitalo:
@mattash Good catch — added `hub/management/commands/backfill_reading_sequence.py` in f0fd0e4 to address this.

It's a dedicated, idempotent command (kept separate from the migration, per your suggestion):
- Recomputes the engine's reading order per `Day` (optionally scoped by `--church`/`--start-date`/`--end-date`) and matches existing `Reading` rows by `(book, start_chapter, start_verse, end_chapter, end_verse)` — the same fields as `unique_reading_per_day` — updating only `sequence`. It never creates or deletes rows.
- `--dry-run` reports pending updates without saving.
- `--strict` exits nonzero if any `Day` has DB rows or engine readings it couldn't match one-to-one (e.g. a book name the engine has since renamed); those are also logged to stderr on every run, strict or not.
- Each `Day`'s updates run in their own `transaction.atomic()`, so the command is restartable and a bad day doesn't roll back the rest.
- Tests cover reversed pre-existing rows, dry-run, idempotency on a second run, unmatched-row reporting (without mutation), `--strict` failure, and date/church scoping.

Also added a section to the PR description with the exact deploy steps. It's a manual step for now (no `release:` process in the Procfile to hook it into) — flagging that in case you'd rather it be automated.

mattash (approved):
looks good!

andylitalo:
Re-targeted to main since original base branch `feat/retire-sacredtradition-readings` has been merged. Merging.

PR #462, opened 2026-07-22
Retire sacredtradition.am for feast names: lectionary engine

andylitalo:
## Summary

Feast-day names were the **last** runtime sacredtradition.am scrape in bahk. This retires it by computing feast names offline from the `armenian_lectionary` engine, mirroring the readings retirement in #461. With engine **1.2.3** it also serves the Armenian (`hy`) feast name offline (1.2.2 fixed feast/book name typos; 1.2.3 adds further reading fixes), so `name_hy` is populated too — closing the one gap the initial (English-only) retirement left, and removing the last reason feast names needed the live scrape.

Stacked on **#461** (`feat/retire-sacredtradition-readings`) — this PR is based on that branch so the diff shows only the feast changes. Merge/rebase after #461.

## Changes

- **`hub/services/feast_service.py`** — `get_feast_for_date(date_obj, church)` reads the engine's `"Liturgical Day"` in both `en` and `hy` and returns the same `{name, name_en, name_hy}` shape `scrape_feast` produced (or `None`). `name_hy` is the engine's Armenian name, left `None` only when the engine has no Armenian form (those keep falling back to English via `name_i18n` and can upgrade later). Filters engine placeholders, unsupported churches, and out-of-range dates (2001–2027).
- **`hub/utils.py`** — removes `scrape_feast`, `_fetch_sacredtradition`, `_stable_url_key`, the circuit-breaker constants, and the now-dead `urllib`/`hashlib`/`sentry_sdk` imports. `get_or_create_feast_for_date` lazily imports the service (avoids a circular import); it already backfills a *missing* `name_hy`, so existing feasts missing Armenian now self-heal.
- **`import_feasts`** management command points at the new service.
- **`requirements.txt`** — `armenian-lectionary >=1.2.3` (feast-name accuracy contract + offline `hy` feast/book names + typo/reading fixes).
- **sacredtradition.am scrub** — updated the now-inaccurate references on active surfaces: `Reading.text_hy_fetched_at` help text (**migration 0056**, chained after the base's `0055` drift-sync; help-text only, no schema change), the admin fieldset label, and the `fetch_armenian_reading_texts` command help/docstring (all compose from the offline `BibleVerse` corpus). Provenance/attribution notes on the corpus and the one-time `scripts/scrape_bible_hy.py` are intentionally retained.
- **Tests** — `test_feast_service.py` updated for the two-language engine call (asserts Armenian populates, and stays `None` when untranslated).

## Scope note: Armenian verse *text*

[...]

PR #468, opened 2026-07-28
Store Scripture text by passage, not by day

andylitalo:
## The problem

Retrieval cost scaled with the `Reading` table, and that table grows ~4 rows a day forever. Every reading row wanted its own text and its own API.Bible call, so the weekly refresh got more expensive every week — the design wanted ~8,600 calls/month and rising against a 5,000/month plan quota. #466 stopped the immediate burn (the refresh retry spiral); this fixes the shape of the cost.

The lectionary is far more repetitive than the reading table suggests. Measured against `armenian_lectionary` v1.2.3: **15,142 reading instances over 2026–2035 resolve to 1,124 distinct passages** — 13.5× dedup — and a single year already covers 1,102 of them. Zero unmapped book names.

## The solution

Store text per **passage**, not per reading or per day.

- `Reading.passage_key` — `"{USFM}.{ch}.{v}-{ch}.{v}"`, derived in `save()` via `hub.constants.passage_key`.
- `PassageText` — one row per `(passage_key, language)`. One table for every language.

Cost now scales with the distinct-passage count, which saturates. Keeping the entire lectionary inside API.Bible's 30-day freshness window costs **~1,270 calls/month, flat forever**, leaving the 5,000 quota with ~3× headroom. The monthly ceiling becomes an alarm rather than a binding constraint.

The invariant worth stating plainly: **never key text by date.** That is the bug this design exists to prevent.

## The refactor it required

- **Fetchers became pure retrieval.** Each takes a citation and returns `{text, version, copyright, fums_token}` or `None`; `store_passage_text` is the single writer. This removed the split where English wrote via `queryset.update()` and Armenian via `save()` into the modeltrans JSON column — two storage conventions that had already drifted apart.
- **Freshness is per `(passage, language)`.** A language absent from `LANGUAGE_TEXT_MAX_AGE_DAYS` never expires, so a locally-composed corpus has no clock — and English arriving from the shared store can no longer suppress Armenian, which is the regression the per-language gate exists to prevent.
- **The refresh task iterates a ≤1,124-row table** per language instead of scanning readings, retiring the nearest-first selection machinery. Locally-composed languages skip the budget, breaker and inter-call delay entirely.
- **`fetch_text_for_reading` delegates** rather than being a second implementation, so the admin button no longer misses the main path's fixes.
- **Adding a language is two steps**: write `fetch_<lang>`, register it in `TEXT_FETCHERS`. No new columns, no modeltrans registration.
- New commands: `warm_passage_texts` (enumerate the corpus once, then no date is ever cold) and `backfill_reading_passage_keys` (escape hatch if key derivation changes).
- Settings: `READING_REFRESH_LIMIT` 2000 → 1500 and now counts **distinct passages**; `READING_FETCH_DAILY_BUDGET` 75 → 25.

### Book names normalise; versification does not

Book *names* collapse onto USFM, so spelling variants share a key — that is language-neutral by construction. Per-edition **versification** deliberately stays out of the key: KJVAIC splits Greek Esther into `ESG 1-7` while the Armenian Nor Ejmiatsin corpus keeps it inline as `EST 10-16`. Baking either mapping into the shared key would impose one edition's quirk on the other, so each fetcher applies its own and the key stays the citation as written.

The last commit widens that Esther remap. It previously handled only chapter 10 — the one citation the engine emits (`Esther 10.4-9`, 29 occurrences, the sole distinct Esther passage in the data). A revision citing Addition C (`Esther 13:8-14:19`) would have asked for `EST 13`, which exists in neither NKJV nor KJVAIC. It is now the general `EST 10-16 → ESG 1-7` rule: chapter 10 is split (10:1-3 close the Hebrew narrative, 10:4 onwards is Addition F), 11-16 are additions entire, verse numbers untouched. A range straddling both warns and degrades to the canonical part rather than raising. Nothing changes on the Armenian side.

### FUMS

Readings citing one passage now share a token. FUMS tracks *displays*, and the app already served one stored token to unlimited users across the 30-day window, with the client deduplicating per device and session. The docstrings asserting a per-row rule are rewritten with that reasoning, and a named test pins the behaviour so it is not reverted by accident.

## Tests

**1,351 run, 1 pre-existing error** — `cleanup_orphaned_bookmarks_async`, which needs a live Celery result backend (the same one #466's description documents). Note it also makes `--parallel` fail outright: that error's traceback cannot be pickled across workers, so run serially.

Rewrites the ~10 tests that encoded the old per-row invariant, and adds coverage for key derivation, book-name variant collapsing, the Esther fetcher/key split (including the chapter-spanning Addition C case and both straddle branches), per-language expiry, one-call-per-passage refresh, zero-call serving of unseen dates, and a query-count test proving the backfill scales with passages rather than rows.

**Migration verification** was run end-to-end rather than argued: built a pre-migration DB with a real year of readings and legacy text, ran `0056`–`0058`, and confirmed 1,512 readings → 1,102 passages, all keyed, with 1,102 English and 1,102 Armenian rows seeded, timestamps and FUMS tokens preserved. The reconciliation report came out exact — engine and stored readings agree — and warming the full four-year corpus afterwards cost only 21 further calls, because the seed had already covered what the DB held.

## Depends on

Stacked, and the base reflects it — this targets `feat/retire-sacredtradition-readings`, not `main`.

- **#461** (`feat/retire-sacredtradition-readings`) — the base. Supplies the lectionary engine and the offline `BibleVerse` corpus this keys against.
- **#466** (`feat/api-bible-spend-control`) — merged into this branch, so its commits appear here until it lands on `main` and #461 picks it up. Review it there, not here.

If #461's review changes `get_daily_readings` or the `BibleVerse` composition, this needs a re-merge.

Sibling PRs on the same base, not dependencies: #465 (Azariah), #462 (feasts), #467 (reading order).

## After merge — deployment

1. **Run migrations.** `0057` backfills passage keys and moves existing text into `PassageText`; it is data-bearing, so let it finish before serving.
2. **Warm the corpus once** — `python manage.py warm_passage_texts --dry-run` first, then for real. After this no date is ever cold. Expect ~1,100 calls on a cold store, far fewer if `0057` seeded from existing text.
3. **Confirm `MAX_READINGS` is unset**, carried over from #466's deploy note.
4. **Watch the dedup ratio** the weekly refresh logs. If it drifts toward 1.0, passage keying has regressed and spend is scaling with the table again — that is the signal to act on.
5. **Alert on `BIBLE_API_MONTHLY_BUDGET` rather than raising it.** Post-change it should never be reached; reaching it means something is wrong upstream.

No env var is required to change. `READING_REFRESH_LIMIT` now counts distinct passages rather than readings, so any environment overriding it should be re-read with that meaning in mind.

## Follow-up (not blocking)

API.Bible access returns **2026-08-04**. Two things to check on that trip: that ESG chapters map to EST 10-16 at a flat offset across all seven (assumed, and inherited from the previous code), and the Azariah 67-vs-68 tail from #465.


mattash (changes requested):
#468 contains an earlier snapshot of #466 and is missing its final budget and telemetry fixes. Please wait for #466 to merge, then rebase this branch onto updated `main` and remove the duplicated #466 changes from this PR's diff.

During conflict resolution, preserve #466's current monthly-budget coverage and telemetry. Adapt those guarantees to the passage-based design:

- Admin single and bulk fetches must respect the monthly API.Bible ceiling.
- Bulk fetches must deduplicate by `passage_key`, making at most one request per distinct passage.
- Add regression tests proving an exhausted budget makes zero API calls and duplicate readings make one call per passage.

Then remove the two unused imports and get exact-head CI green.

Because this PR contains schema/data migrations, merge also requires the repository's hold-lane approval: database snapshot, seeded/keyed row-count verification, and a defined rollback window.

andylitalo:
Addressed. Point by point:

**"#468 contains an earlier snapshot of #466."** It no longer does — the branch was retargeted at `feat/api-bible-spend-control` and merged up to that branch's current head (`8b56f2b`), which includes `9ea0961` ("Enforce API.Bible budget on every fetch path, fix refresh telemetry"). Since #466 is the PR base, GitHub computes this diff from `8b56f2b`, so #466's changes are not in it: the four commits shown are this PR's own. `origin/main` is also an ancestor, so there is nothing to rebase onto — when #466 lands, the base retargets to `main` with no diff change. I have left #466 to merge on its own rather than merging it myself.

**Admin single and bulk fetches respect the monthly ceiling.** Both already route through `fetch_all_reading_texts(..., budgets=bible_api_budgets(include_daily=False))`; that guarantee survived the passage-based rewrite, and `test_single_fetch_view_refuses_when_monthly_budget_exhausted` / `test_bulk_fetch_action_stops_at_monthly_budget` carried over with it. Added `test_bulk_fetch_makes_zero_calls_when_budget_exhausted` to pin the stronger claim you asked for: an exhausted ceiling makes *zero* calls, not zero-after-the-first.

**Bulk fetches deduplicate by `passage_key`.** This was a real gap — the one retrieval path still iterating rows. The refresh task, `warm_passage_texts`, `fetch_reading_texts` and the readings view all collapse onto distinct keys first; the admin bulk action made one call per selected reading. Fixed in `3bb150a`: group by `passage_key`, one fetch per distinct passage, and readings whose book has no USFM mapping are reported rather than sent at an API that cannot address them. The Armenian bulk action gets the same grouping — local composition, no quota at stake, but one task per row recomposed the same passage once per date citing it.

**Regression tests.** `test_bulk_fetch_makes_one_call_per_distinct_passage` (4 readings, 2 distinct passages, one pair differing only in book-name spelling → 2 calls, all 4 served), `test_bulk_fetch_makes_zero_calls_when_budget_exhausted`, `test_bulk_armenian_fetch_enqueues_one_task_per_passage`, `test_bulk_fetch_skips_unmappable_readings_without_calling`. I checked the two dedup tests fail on the pre-fix admin code (4 != 2 and 3 != 2) rather than passing vacuously.

**Unused imports.** One remained on exact head — `hub.constants.passage_key` in `reading_text_service.py`; the other was already fixed in #466's `8b56f2b`. `ruff check .` is clean.

**Full suite:** 1,289 tests, 1 error — `test_cleanup_task_deletes_all_orphans_across_batches`, pre-existing and environmental (it needs the Celery result backend at `redis:6379`, which my local run has no route to; CI's redis service provides it). Nothing else fails.

**Hold-lane approval**: I'll load readings into a test db and test the migration before we merge.

mattash (approved):
Approved. I conducted an isolated PostgreSQL migration rehearsal against this exact head and its base. Migrations 0057–0059 completed successfully; representative duplicate, timestamp, translation, and unmappable cases all passed, with legacy rows and metadata preserved by hash.

andylitalo:
Thank you @mattash ! Merging.

PR #469, opened 2026-07-29
Stitch NKJV Romans doxology (16:25-27) for Armenian-numbered 14:24-26 readings

andylitalo:
## Goal

Document and patch a specific versification discrepancy: the Armenian liturgical lectionary numbers the closing Romans doxology as part of chapter 14 (e.g. `"13.11-14.26"`), but NKJV ends chapter 14 at verse 23 and places the same doxology text at 16:25-27. Before this fix, fetching NKJV text for such a reading would ask API.Bible for verses (`ROM.14.24`-`ROM.14.26`) that don't exist in NKJV's numbering, silently failing and leaving `Reading.text` empty while the Armenian side (sourced from the offline Nor Ejmiatsin corpus, which does number the doxology within chapter 14) populated fine.

**Stacked on #467** (`feat/reading-sequence-order`) — based on that branch since it depends on the `reading_text_service.py` / `fetch_english_text` refactor introduced there.

## What changed

- `BibleAPIService.resolve_reading_segments()`: resolves a Reading into one or more API.Bible passage segments. Romans citations reaching past NKJV's 14:23 split into `[main text through 14:23, doxology at 16:25-27]` (with correct partial-range handling). Everything else (including the existing Esther/`ESG` remap) still resolves to a single segment.
- `BibleAPIService.get_composite_passage()`: fetches each segment and concatenates their content. For multi-segment reads, only the **first segment's** FUMS token/version/copyright is kept — API.Bible issues one token per call, and the primary segment is the bulk of what's read.
- `fetch_text_for_reading` (admin) and `fetch_english_text` (Celery refresh, readings view) both now route through the new segment-based resolution.

## Why this may not be the final shape

This is a targeted, one-off fix for a single known book/reading. There may be other liturgical-vs-NKJV versification mismatches lurking elsewhere that would hit the same silent-failure mode. Given the Daniel 3 / Prayer of Azariah composite-reading work already in this codebase (deriving a standalone `S3Y` book from the Armenian corpus), a more general "composite/versification override" layer might be worth building instead of accumulating book-specific special cases in `resolve_reading_segments`. Opening this now mainly to close the immediate gap and put the discrepancy on record — happy to fold it into a broader abstraction if/when more cases turn up.

## Test plan

- [x] `hub/tests/test_bible_api.py` — 6 new tests: segment resolution (full/partial/doxology-only/in-range/non-Romans), passage stitching (single + multi-segment), and an end-to-end refresh-task test confirming NKJV text is `[11]...[23]... [25]...[27]` for Romans 13.11-14.26.
- [x] Full `test_bible_api.py` suite (58 tests) passes.
- [x] Related reading/lectionary suites (`test_import_readings`, `test_reading_context`, `test_lectionary_service`, `test_armenian_text`; 41 tests) pass unaffected.

### DIFF
--- a/hub/services/bible_api_service.py
+++ b/hub/services/bible_api_service.py
@@ -22,6 +22,15 @@ BASE_URL = "https://rest.api.bible/v1"
 NKJV_BIBLE_ID = "63097d2a0a2f7db3-01"
 KJVAIC_BIBLE_ID = "a6aee10bb058511c-01"  # KJV with Apocrypha, American Edition
 
+# The Armenian liturgical tradition numbers the closing doxology of Romans as part
+# of chapter 14 (verses 24-26). NKJV -- like most Western Bibles -- ends chapter 14
+# at verse 23 and places the same doxology at 16:25-27. Citations that reach past
+# 14:23 are split into two segments (main text + doxology) and stitched back
+# together; see resolve_reading_segments / get_composite_passage.
+ROMANS_CH14_LAST_VERSE = 23
+ROMANS_DOXOLOGY_CHAPTER = 16
+ROMANS_DOXOLOGY_START_VERSE = 25
+
 
 class BibleAPIService:
     """Client for extracting verses from API.Bible.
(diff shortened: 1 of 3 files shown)


andylitalo:
Replacing with more general solution.

PR #471, opened 2026-08-01
Widen Feast.name to 512 characters (migration hold lane)

andylitalo:
@mattash I'm fine with leaving at 256 and cutting off the end of the feast names, but I figured it would be better to have the full name in the backend and decide how much to present on the frontend.

## Summary

Two Armenian lectionary feast names enumerate their saints inside the name and exceed the 256-character `Feast.name` column:

| name | length | dates affected |
|---|---:|---:|
| `The Twelve Holy Doctors of Church: Hierotheus of Athens, …` | 289 | 27 |
| `The Holy Fathers of Egypt: Paul the Hermit, …` | 257 | 27 |

Both recur annually, so **54 dates in 2001–2027** are affected. On PostgreSQL the insert raises `DataError`: the API degrades to "no feast" for that day, and a range import aborts partway through.

## This is pre-existing, not a regression

The strings are **byte-identical** to what sacredtradition.am served, so the retired `scrape_feast` hit exactly the same error on exactly the same days. Surfaced during the [#462](https://github.com/surp-hovhannes/bahk/pull/462) review, but it predates that work.

Widened rather than truncated because the names are correct — just long. Truncating would cut mid-enumeration (`…John Chrysostom, and Cyri…`).

## Why no test caught it

The test database is **SQLite, which does not enforce `max_length`** — it stores an over-long value happily and the assertion passes, while PostgreSQL raises. The new tests therefore go through `full_clean()`, which applies Django's validators on any backend.

## Forward safety

On PostgreSQL, increasing a `varchar` length limit is a **catalog-only change**: no table rewrite, no full-table lock, no data touched. Safe to apply while serving traffic.

## Rollback

The reverse migration is **not** safe once a stored name exceeds 256 — PostgreSQL refuses to narrow a column whose data would not fit. Documented down-path:

```sql
-- capture first; the truncation is lossy and irreversible
SELECT id, day_id, name FROM hub_feast WHERE LENGTH(name) > 256;

UPDATE hub_feast SET name = LEFT(name, 256) WHERE LENGTH(name) > 256;
```
```bash
python manage.py migrate hub 0059
```

`Feast.name_hy` is unaffected either way — django-modeltrans stores it in the `i18n` JSONField, which has no length limit.

## Changes

- `hub/models.py` — `Feast.name` `max_length` 256 → 512, with the reason inline.
- `hub/migrations/0060_alter_feast_name.py` — the `AlterField`, carrying the forward-safety and rollback notes in its docstring.
- `hub/tests/test_feast_name_storage.py` — four regression tests using the verbatim over-limit names, asserting via `full_clean()` so SQLite cannot mask the failure, and checking that widening did not make the column accept anything.

## Verification

```bash
python manage.py test hub.tests.test_feast_name_storage tests.unit.hub --settings=tests.test_settings
# Ran 53 tests — OK
python manage.py makemigrations --check --dry-run --settings=tests.test_settings
# hub clean
```

Note: `makemigrations --check` reports pending migrations for the **`icons`** app. That is pre-existing drift on `main`, unrelated to this change, and deliberately left alone.


mattash:
The migration itself looks correct after the renumbering, but the PR description still contains the old migration numbers. This is now `0060_alter_feast_name`, so the rollback target should be `hub 0059`, not `hub 0056`. Running the documented command would also reverse migrations 0057–0059, including the `PassageText` migration. Could you update both the Changes section and rollback example to match the final migration numbering?

andylitalo:
Updated — Changes section now points to `0060_alter_feast_name.py` and the rollback command targets `hub 0059`, matching the final migration numbering after the rebase.

mattash (approved):
💯

PR #472, opened 2026-08-01
Exercise the real lectionary engine in feast_service tests

andylitalo:
Addresses the review on **#462**, on which this is stacked (base = `feat/retire-sacredtradition-feasts`). Merge after #462.

## The gap the review found

`hub/tests/test_feast_service.py` mocks `compute_armenian_lectionary`. That is right for testing this module's own branching, but it means **no test ever ran the wheel that ships**. Every defect the review found lived in that gap:

- `(movable ordinary-time reading)` on 2011-02-04/06/09/11 and 2022-02-04 → service returned `None`, so the day silently lost its feast;
- a wrong name on 2011-02-13, with no `name_hy`;
- 54 dates whose name overflows `Feast.name`.

## Real-engine coverage

`hub/tests/test_feast_service_real_engine.py` calls the actual engine across the whole supported **2001–2027** window. It is offline, pure-Python and in-process, so ~9.9k dates cost about a second.

Two details are what let it catch things the mocked tests could not:

1. **It goes through `full_clean()`** — the test DB is SQLite, which does not enforce `max_length`. A bare `save()` stores an over-long name happily while PostgreSQL raises `DataError`. `full_clean()` applies Django's validators on any backend.
2. **It asserts on `name_hy`** — a missing Armenian name is the cheapest signal that the engine invented an English name rather than serving one the source uses. All six review-flagged dates had `name_hy = None` for exactly that reason.

The sweep also covers **2027**, which no ground-truth test can reach: sacredtradition.am publishes nothing for that year, so the engine's own oracle tests stop at 2026.

Explicit regression cases for the review's dates, asserting the corrected names in both languages:

| date | name | Armenian |
|---|---|---|
| 2011-02-04 / 02-09 / 02-11, 2022-02-04 | `Fast day` | `Պահք` |
| 2011-02-06 | `Fourth Sunday after Nativity` | `Դ կիւրակէ զկնի Ս. Ծննդեան` |
| 2011-02-13 | `Fifth Sunday after Nativity` | `Ե կիւրակէ զկնի Ս. Ծննդեան` |

## Service changes

- **`_fit_to_storage`** reads the limit from `Feast._meta.get_field("name").max_length` rather than hard-coding it, so it cannot drift from the column after a future migration. Logs at error level before clamping. It should never fire once the column is widened — it exists because the failure it prevents is disproportionate.
- **The placeholder guard now logs at error level.** Engine ≥1.3 contracts every supported date to a real name (its `test_feast_contract` asserts that across 2001–2027), so reaching that branch means an engine regression, not a normal outcome. Kept as defence in depth: persisting `"(commemoration)"` would be worse than recording no feast.

## Dependencies

Two prerequisites, neither blocking review:

- **`armenian-lectionary` 1.3.0** ([andylitalo/armenian-lectionary#15](https://github.com/andylitalo/armenian-lectionary/pull/15)) — fixes the engine-side defects. Not released yet, so `requirements.txt` still pins `>=1.2.3` with a note to raise it once tagged. Tests here were run against a local build of that branch.
- **#471** (`Feast.name` 256 → 512) — `test_long_names_are_stored_in_full` **skips** while the column is still 256 and activates by itself once #471 lands and this branch rebases onto it.

## Verification

```bash
PYTHONPATH=<local armenian-lectionary build> \
  python manage.py test hub.tests tests.unit.hub --settings=tests.test_settings
# Ran 483 tests — OK (skipped=1, the one gated on #471)
```


hub/tests/test_feast_service_real_engine.py
```diff
@@ -0,0 +1,129 @@
...
+        checked = 0
+        while date_obj <= end:
+            result = get_feast_for_date(date_obj, self.church)
+            checked += 1
+            if result is None:
+                missing.append(date_obj.isoformat())
+            else:
+                if result["name_hy"] is None:
+                    no_armenian.append(date_obj.isoformat())
+                if len(result["name"]) > max_length:
```
mattash: `result["name"]` has already passed through `_fit_to_storage`, so this condition cannot observe an overflow—the service will always return a value at or below `max_length`. Could the sweep inspect the engine's raw English `Liturgical Day` before calling the service? While the field is still 256, the assertion can allow exactly the two known long names; after #471 widens it, the expected overflow set becomes empty. That preserves the current sequencing while ensuring a new engine name cannot be silently truncated without failing the contract test.

hub/services/feast_service.py
```diff
@@ -67,8 +95,18 @@ def get_feast_for_date(date_obj, church) -> dict | None:
...
+    if any(marker in name_en for marker in _NON_FEAST_MARKERS):
+        logger.error(
+            "Lectionary engine returned the placeholder %r for %s; recording no feast. "
+            "The engine is contracted to serve a real name on every supported date, so "
+            "this indicates an engine regression.",
+            name_en, date_obj,
+        )
+        return None
+
+    name_en = _fit_to_storage(name_en, date_obj)
```
mattash: There is a subtle interaction between truncation and untranslated-name detection here. If an over-limit feast has no Armenian translation, the engine returns the same full English value for `hy`; after `name_en` is truncated, the later equality check fails and the full English string is treated as Armenian. Could we retain `raw_name_en`, compare `name_hy` against it, and only then call `_fit_to_storage` for the stored English value? A regression test with an untranslated `max_length + 1` name should assert that English is clamped while `name_hy` remains `None`.

mattash (commented):
Two narrow fixes to make the storage guard testable and keep untranslated fallback correct; neither changes the assumed 1.3.0 rollout sequencing.

andylitalo:
Rebased onto #480 (the feast re-key) and retargeted; conflicts resolved, `MERGEABLE`/`CLEAN`.

**Kept as-is** — none of this is superseded by the re-key, which changes the model key while this
hardens the service:
- `_fit_to_storage`, reading the limit from the field so it can't drift from the column
- error-level logging when the engine returns a placeholder, instead of silently recording no feast
- the real-engine contract tests

**Changed in the rebase:**

1. Tests build feasts on `church` rather than `day` — `Feast` is keyed by `(church, name)` now.
2. The `skipIf` on `test_long_names_are_stored_in_full` is **gone**. It was waiting on the
   widening (#471), which has since landed on main, so the test now runs unconditionally.
3. `requirements.txt` keeps the pin at `>=1.3.0` and the explanation of *why* 1.3.0 is required,
   but drops the `NOTE: 1.3.0 is not released yet` line — it is released and on PyPI.
4. **One fixture was wrong against the shipped engine.** `2011-02-13` is
   `"Fifth Sunday after Nativity — Eve of Fast of Catechumens"`, not the bare Sunday name. I
   checked rather than just relaxing the assertion: the Fast of the Catechumens opens Monday
   `2011-02-14`, and released 1.3.0 labels the day before a fast begins. The fixture was written
   before that shipped, so the engine is right and the expectation was stale.

Full suite: **1417 passing.** The one error, `test_bookmark_tasks`, is a pre-existing Redis
connection failure that fails identically on clean `main`.

Pre-rebase tip is preserved locally as `backup/472-pre-rebase` (`dc35532`).

andylitalo:
Closing without merging. The truncation clamp (_fit_to_storage) and its real-engine sweep test existed to guard against a feast name overflowing Feast.name (512 chars). Checked the actual numbers: the longest name anywhere in armenian_lectionary's shipped data is 289 chars, and the worst-case theoretical compounding with a position-label prefix and eve-label suffix tops out at 417 — comfortably under 512.

The engine's own runaway-concatenation guard (MAX_PLAUSIBLE_NAME) was a loose 600 (2x the observed 289 at the time, no other justification). Tightened that to an evidence-based 450 upstream instead: andylitalo/armenian-lectionary#23. With that in place, this bahk-side defensive clamp is redundant — closing rather than merging it.

Issue #474, opened 2026-08-09
Readings from API.Bible not fetched


andylitalo:
**Currently blocking the ability to serve English readings to the frontend and landing page.**

Clicking "Re-fetch Bible text" for English text on admin page leads to error message: `Failed to fetch Bible text for Isaiah: Chapter 7, Verses 1-9. Check logs for details.` Try at `https://api.fastandpray.app/admin/hub/reading/2923/change/` for example.

Requests confirmed to have reached API.Bible and API.Bible request limit of 5k not reached.

<img width="1357" height="297" alt="Image" src="https://github.com/user-attachments/assets/f65c76c2-2b56-476a-993d-814ef978e6d7" />

andylitalo:
Resolved once request budget refreshed at the end of the monthly billing cycle. New approach of storing one db entry per reading instead of per reading per day should prevent exhaustion of our budget in the future. Closing this issue.

PR #476, opened 2026-08-09
Add per-edition circuit breaker for API.Bible fetches

andylitalo:
## Summary
- Adds a per-edition (NKJV vs KJVAIC) circuit breaker (`hub/services/circuit_breaker.py`) that opens after `BIBLE_API_CIRCUIT_BREAKER_THRESHOLD` (default 3) consecutive failures and skips further attempts for `BIBLE_API_CIRCUIT_BREAKER_COOLDOWN_SECONDS` (default 900s) without charging budget
- Fixes a starvation bug found while investigating #474: a rejected NKJV entitlement was burning the whole `READING_FETCH_DAILY_BUDGET` (25/day) on doomed retries, starving KJVAIC (Apocrypha) fetches that share the same budget and would otherwise succeed
- Split out of #475 — that PR's `warm_passage_texts` fix is independent and already unblocks the immediate issue (NKJV access itself is now confirmed working again), so this is general hardening rather than something blocking the current fix
- New tests: `hub/tests/test_circuit_breaker.py` (unit) + `FetchEnglishCircuitBreakerTests` in `hub/tests/test_bible_api.py`, including the test pinning the actual bug (stuck NKJV must not block working KJVAIC)

## Test plan
- [x] `tests.unit.hub`, `hub.tests.test_bible_api`, `hub.tests.test_circuit_breaker`, `hub.tests.test_armenian_text`, `hub.tests.test_task_imports` — 184 tests, all passing (native venv, no Docker)
- [x] Manual repro against live API.Bible: breaker stays closed on real NKJV/KJVAIC success

### DIFF
--- a/hub/services/circuit_breaker.py
+++ b/hub/services/circuit_breaker.py
@@ -1,0 +1,99 @@
+"""Per-source circuit breaker so a persistent failure cannot drain a shared spend budget.
+
+``APIBudget`` (see ``hub/services/api_budget.py``) caps *total* spend; it says nothing
+about whether calls are succeeding.  Without this, a source that is rejecting every
+call (bad credentials, a missing entitlement, an outage) still burns one budget slot
+per attempt, and a small shared daily budget can be exhausted entirely on doomed
+retries before an unrelated, healthy source gets a turn.  This stops attempts to a
+source once it has failed ``threshold`` times in a row, until ``cooldown_seconds`` has
+passed -- so a rejected source stops spending budget quickly, and self-heals once
+whatever was wrong is fixed, without needing a manual reset.
+"""
+
+import logging
+
+from django.core.cache import cache
+
+logger = logging.getLogger(__name__)
+
+
+class ConsecutiveFailureBreaker:
+    """Opens after *threshold* consecutive failures for one *name*; auto-closes after
+    *cooldown_seconds*.
+
+    Backed by ``django.core.cache`` (Redis in production), so state is shared across
+    web dynos and Celery workers -- unlike a local loop variable, this catches failures
+    that accumulate across separate requests, which is exactly the case a shared daily
+    budget needs protecting from.
+
+    Fails *open* (permits the attempt) if the cache is unreachable.  This differs from
+    ``APIBudget``, which fails closed: a budget's job is to guarantee a ceiling is never
+    exceeded, so it must refuse spend it cannot verify.  A breaker's job is only to stop
+    wasting spend on a source known to be failing; if the cache itself is down, calls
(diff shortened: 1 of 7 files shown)


andylitalo:
This is probably overkill but Claude found this potential bug if one of the Bible versions is not available. Not intending to merge but want to record and consider before I close.

mattash (commented):
Two fixes would make this circuit breaker safe to rely on if the PR is revived. They are recorded inline; neither affects the already-deployed #475 backfill fix.

hub/services/reading_text_service.py
```diff
@@ -159,12 +172,14 @@ def fetch_english(
     try:
         result = service.get_passage(*passage)
     except Exception as exc:
+        breaker.record_failure()
```
mattash, line 175: This treats every `get_passage` exception as an edition-wide API failure. But `resolve_reading_passage` deliberately leaves verse validation to API.Bible, where an invalid citation returns 404. Three malformed readings would therefore open the NKJV/KJVAIC breaker and withhold otherwise healthy passages for the cooldown.

Only count failures that indicate a source-wide problem: connection/timeouts (no HTTP response) and 401/403/429/5xx responses. Do not increment the breaker for malformed-input responses such as 400/404; log and surface those as data errors instead. Please add a test where three 404s leave the breaker closed and a following valid passage for that edition is fetched successfully.

hub/services/circuit_breaker.py
```diff
@@ -0,0 +1,99 @@
...
+            return False
+
+    def record_failure(self) -> None:
+        """Count one failure; open the breaker once *threshold* is reached in a row."""
+        try:
+            try:
+                count = cache.incr(self._failures_key())
+            except ValueError:
+                # Key missing or expired: this is the first failure of a new streak.
+                cache.set(self._failures_key(), 1, timeout=self.cooldown_seconds)
```
mattash, line 66: `cache.set` is unsafe after a missing-key `incr` under concurrent requests: multiple workers can all receive `ValueError` and then each overwrite the counter with `1`. The breaker may then stay closed while the same burst consumes the shared daily budget.

Use the existing `APIBudget._increment` pattern: `cache.add(..., timeout=...)` to initialize without overwriting another worker; if it returns false, retry `cache.incr` (handling the tiny expiry race). Add a deterministic mocked test for `incr -> ValueError`, `add -> False`, then a successful retrying `incr`.

andylitalo:
Closing because the issue has been fixed on API.Bible (requested NKJV as a Bible on our account--was dropped while transitioning accounts). Could be helpful in the future but adds too much churn in light of other more important efforts like replacing sacredtradition.am with armenian-lectionary engine.

PR #477, opened 2026-08-12
Expose context_eligible on GetFeastForDate response

andylitalo:
## Summary
- `GetFeastForDate` already computes `should_trigger_generation = is_feast_context_generation_eligible(feast)` (`hub/views/feasts.py:129`) to decide whether to enqueue AI feast-context generation, but never returned that boolean to API clients.
- The `bahk_react_native` frontend was instead re-deriving eligibility client-side from the raw `designation` field (`designation === 'Fast'`). That field is set by a **one-shot** async LLM classification task (`hub/signals.py` → `determine_feast_designation_task`) with no retry on failure, and is frequently `null` in production — including for the exact feast that surfaced this bug, and for a year-old example that never got classified. So the frontend check silently failed to hide AI context on many generic fast days, trapping users in a permanent "Context is being generated. Please refresh." loop.
- This PR just exposes the already-computed, more-robust `context_eligible` boolean in the response (`feast_data["context_eligible"]`) so the frontend can gate on it directly instead of guessing from `designation`. No change to `is_feast_context_generation_eligible()` itself.

This is stage 1 of a 3-stage fix (companion frontend PR in `bahk_react_native` #422 consumes this field). Stage 2/3 replace the underlying LLM+regex eligibility heuristic with a deterministic calendar-based one (`armenian-lectionary` PR #16) — this PR's `context_eligible` field is deliberately an outcome contract, so that future swap won't require any frontend changes.

## Test plan
- [x] Extended the 6 existing `hub/tests/test_feast_views.py` eligibility-scenario tests (`test_api_route_{enqueues,does_not_enqueue}_context_for_*`) with matching `context_eligible` assertions, rather than adding new tests — each already sets up an eligible/ineligible fixture and asserts on `generate_feast_context_task` call behavior.
- [x] `./venv/bin/python manage.py test hub.tests.test_feast_views --settings=tests.test_settings` — all 35 tests pass
- [x] Confirmed the two full-response-dict-equality tests (no-feast path) are unaffected — they return before `feast_data` is built


mattash (approved):
Reviewed against current main after #482; merge is clean and the full validation suite passes.

andylitalo:
Thanks! Merging.

